Issue #2, opened 2018-07-26
Promotion rules


oli-obk:
Promotion is the process of allowing `let x: &'static u32 = &1;` to be valid in Rust. Technically the `&1` produces a temporary local variable and references to that don't have the `'static` lifetime. But there is a check which figures out expressions that are constant and then another pass inserts an unnameable static for the value and refers to that static instead.

This was introduced in https://github.com/rust-lang/rfcs/pull/1414 and I'm doing a write up for it in https://github.com/rust-rfcs/const-eval/blob/master/promotion.md

Now there are some headaches with promotion. E.g.

```rust
let x: &'static usize = &(0 - 1);
```

which, while being const evaluable, would cause a const eval error due to overflow.
This is fine in the above case, since the user obviously requested promotion by setting the `'static` lifetime, but

```rust
let x: &usize = &(0 - 1);
```

did not and would still get promoted and then error. We work around this by leaving in the debug assertions that the compiler has on integer arithmetic, which mean even though we promoted the computation to a static, all the checks are still done at runtime (unless optimized out due to guaranteed uselessness).

So everything is fine in these simple cases, it gets problematic when there's code that can't be evaluated at compile-time but which would also *not* panic at runtime. E.g.

```rust
union Foo { x: &'static i32, y: usize }
let x: &bool = unsafe { Foo { x: &1 }.y == Foo { x: &2 }.y };
```

This errors during const eval because const eval does not compare pointers. We could do pointer comparisons, and in this case it would even be fine because `&1` and `&2` can never be the same address, but

```rust
union Foo { x: &'static i32, y: usize }
let x: &bool = unsafe { Foo { x: &1 }.y == Foo { x: &1 }.y };
```

is not so clear-cut. LLVM might decide to put both `&1` and the other `&1` into the same static. Then the addresses would be the same. Or it would not do that, then the addresses are different. When we move to other operators, the result isn't even decideable by LLVM anymore (only the linker knows this):

```rust
union Foo { x: &'static i32, y: usize }
let x: &bool = unsafe { Foo { x: &1 }.y < Foo { x: &2 }.y };
```

So, the solution was to forbid promoting unions (since pointer to usize casts are already not promotable). But this just shifts the problem

```rust
union Foo { x: &'static i32, y: usize }
const A: usize = unsafe { Foo { x: &1 }.y };
const B: usize = unsafe { Foo { x: &2 }.y };
let x: &bool = &(A < B);
```

[...]

PR #6, opened 2018-08-29
Rules for const safety and promotion

RalfJung:
Based on my blog posts and lots of prior discussion with @oli-obk and @eddyb.  I am sure I forgot some of that discussion though, so please let me know what is wrong or missing.^^

### DIFF
--- a/const_safety.md
+++ b/const_safety.md
@@ -1,1 +1,134 @@
 # Const safety
+
+The miri engine, which is used to execute code at compile time, can fail in
+four possible ways:
+
+* The program performs an unsupported operation (e.g., calling an unimplemented
+  intrinsics, or doing an operation that would observe the integer address of a
+  pointer).
+* The program causes undefined behavior (e.g., dereferencing an out-of-bounds
+  pointer).
+* The program panics (e.g., a failed bounds check).
+* The program loops forever, and this is detected by the loop detector.
+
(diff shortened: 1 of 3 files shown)


RalfJung:
I significantly expanded the discussion of promotion, also covering the concerns raised in https://github.com/rust-rfcs/const-eval/issues/2.

const_safety.md
```diff
@@ -1 +1,134 @@
...
+The miri engine, which is used to execute code at compile time, can fail in
+four possible ways:
+
+* The program performs an unsupported operation (e.g., calling an unimplemented
+  intrinsics, or doing an operation that would observe the integer address of a
+  pointer).
+* The program causes undefined behavior (e.g., dereferencing an out-of-bounds
+  pointer).
+* The program panics (e.g., a failed bounds check).
+* The program loops forever, and this is detected by the loop detector.
```
oli-obk: On a best effort basis. We basically don't care too much about nontermination except for diagnostics

RalfJung: AFAIK we do error hard if we detect a reoccurring state? This could happen in a promoted as well.

RalfJung:
Note that while the `promoted.md` contains some TODOs, I think we don't have to wait merging this until the TODOs are figured out -- it's still strictly more informative than what we have now. ;)

Issue #1, opened 2018-07-21
`const fn` and generics


RalfJung:
Now that I came to like const soundness, of course I had to start thinking about how to achieve const soundness in the presence of generics. The problem is that in a `const fn foo<T: Trait>`, when we call `T::method`, how can we know if that method is safe to call in const context? When type-checking the method body, we cannot know if that will later resolve to a `const fn` or not.

There was plenty of discussion on this that I did not follow closely, but it seems it mostly happened in chat so I don't know of a good summary. @Centril was involved, probably @eddyb and @oli-obk and I am not sure who else?

I propose to discuss this further here so that we have a place to look at. I see two fundamental approaches:
* Find some way to solve this and obtain a const sound generic type system, i.e., obtain a static guarantee that (safe) const code will not even attempt to call a non-`const` fn. (By "static" I mean that we detect this before even starting CTFE, just on the type system level. In our context here, "dynamic" means "during CTFE evaluation", i.e. when miri runs).
* Punt on this. We *will* detect this a CTFE evaluation time. Our type system would only be const sound for non-generic code. Not very satisfying, but it's worth keeping in mind that this is not a *horrible* alternative, which puts an upper bound on the amount of complexity we should pile up to avoid this outcome. :D

RalfJung:
Here is the simplest possible thing I could come up with, and I hope @centril et al will tell me why that is not sufficient. ;)

I propose that we have `const impl Trait for Foo` where all methods are checked in const context; and that, whenever we perform trait resolution in const context, we only consider `const impl`. We also freely allow using trait bounds that were introduced by `where` clauses, relying on the fact that whoever is satisfying those `where` clauses will take care to only use `const impl`. And I think that's already it.

So, for example, in
```rust
const fn ptr_eq<T>(x: &T, y: &T) -> bool {
    unsafe { x as *const _ == y as *const _ }
}
```
we check that we have a `const impl PartialEq for *const T`.

In
```rust
const fn foo<T: PartialOrd>(x: &T, y: &T) -> bool {
    x < y || y < x
}
```
we don't actually check anything locally, because the `where` clause is sufficient. We rely on the fact that when `foo` is *called* in const context, the caller will only use `const impl` to satisfy our `where` clause. That way, we should have an invariant that when a trait method is executed in const context, it always comes from a `const impl`.

---

[...]

PR #8, opened 2018-10-05
RFC for trait bounds on generic parameters of const fns

oli-obk:
[Rendered](https://github.com/rust-rfcs/const-eval/blob/generic_const_fn_bounds/const-generic-const-fn-bounds.md)

cc @Centril @RalfJung @eddyb

Once we have gone through the process here, I'll post the final result on the regular RFC repo

Previous discussion: https://github.com/rust-rfcs/const-eval/issues/1

Glossary:

* [RPIT](https://github.com/rust-lang/rust/issues/34511) (return position impl trait)
* [const context](https://rust-lang-nursery.github.io/rustc-guide/const-eval.html)

### DIFF
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -8,8 +8,9 @@
 
 Allow `const impl`s for trait impls where all method impls are checked as const fn.
 
-Make it legal to declare trait bounds on generic parameters of `const fn` and allow
-the body of the const fn to call methods on these generic parameters.
+Make it legal to declare trait bounds on generic parameters of const functions and allow
+the body of the const fn to call methods on the generic parameters that have a `const` modifier
+on their bound.
 
 # Motivation
 [motivation]: #motivation
@@ -21,26 +22,26 @@ generic parameter type), because they are fully unconstrained.
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-You can call call methods of generic parameters of `const fn`, because they are implicitly assumed to be
-`const fn`. For example, even though the `Add` trait declaration does not contain any mention of `const`,
+You can call call methods of generic parameters of a const function, because they are implicitly assumed to be
+`const fn`. For example, the `Add` trait declaration has an additional `const` before the trait name, so
 you can use it as a trait bound on your generic parameters:
 
 ```rust
-const fn triple_add<T: Add>(a: T, b: T, c: T) -> T {
+const fn triple_add<T: const Add>(a: T, b: T, c: T) -> T {
     a + b + c
 }
 ```
 
 The obligation is passed to the caller of your `triple_add` function to supply a type whose `Add` impl is fully
 `const`. Since `Add` only has `add` as a method, in this case one only needs to ensure that the `add` method is
-`const fn`. Instead of adding a `const` modifier to all methods of a trait impl, the modifier is added to the entire
+`const`. Instead of adding a `const` modifier to all methods of a trait impl, the modifier is added to the entire
 `impl` block:
 
 ```rust
 struct MyInt(i8);
 const impl Add for MyInt {
     fn add(self, other: Self) -> Self {
-        MyInt(self.0, other.0)
+        MyInt(self.0 + other.0)
     }
 }
 ```
@@ -65,23 +66,17 @@ const impl Hash for MyInt {
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-This is the technical portion of the RFC. Explain the design in sufficient detail that:
-
-- Its interaction with other features is clear.
-- It is reasonably clear how the feature would be implemented.
-- Corner cases are dissected by example.
-
-The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+The implementation of this RFC is (in contrast to some of its alternatives) mostly
+changes around the syntax of the language (adding `const` modifiers in a few places)
+and ensuring that lowering to HIR and MIR keeps track of that.
+The miri engine already fully supports calling methods on generic
+bounds, there's just no way of declaring them.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-One cannot add trait bounds to `const fn` without them automatically
-requiring `const impl`s for all monomorphizations. Even if one does not
-call any method on the generic parameter, the methods are still required to be constant.
-
 It is not a fully general design that supports every possible use case,
-but only covers the most common cases. See also the alternatives.
+but it covers the most common cases. See also the alternatives.
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
@@ -97,13 +92,17 @@ and this RFC is forward compatible to have its background impl be an effect syst
 One could annotate methods instead of impls, allowing just marking some method impls
 as const fn. This would require some sort of "const bounds" in generic functions that
 can be applied to specific methods. E.g. `where <T as Add>::add: const` or something of
-the sort.
+the sort. This design is more complex than the current one and we'd probably want the
+current one as sugar anyway
 
-## Explicit `const` bounds
+## No explicit `const` bounds
 
-One could require `T: const Trait` bounds to differentiate between bounds on which methods
-can be called and bounds on which no methods can be called. This can backwards compatibly be
-added as an opt-out via `T: ?const Trait` if the desire for such differences is strong.
+One could require no `const` on the bounds (e.g. `T: Trait`) and assume constness for all
+bounds. An opt-out via `T: ?const Trait` would then allow declaring bounds that cannot be
+used for calling methods. This design causes discrepancies with `const fn` pointers as
+arguments (where the constness would be needed, as normal function pointers already exist
+as the type of constants). Also it is not forward compatible to allowing `const` trait bounds
+on non-const functions
 
 ## Infer all the things
 
@@ -117,7 +116,40 @@ This is strictly the most powerful and generic variant, but is an enormous backw
 hazard as changing a const fn's body to suddenly call a method that it did not before can break
 users of the function.
 
+# Future work
+
+This design is explicitly forward compatible to all future extensions the author could think
+about. Notable mentions (see also the alternatives section):
+
+* an effect system with a "notconst" effect
+* const trait bounds on non-const functions allowing the use of the generic parameter in
+  constant expressions in the body of the function or maybe even for array lenghts in the
+  signature of the function
+* fine grained bounds for single methods and their bounds
+
+It might also be desirable to make the automatic `Fn*` impls on function types and pointers `const`.
+This change should probably go in hand with allowing `const fn` pointers on const functions
+that support being called (in contrast to regular function pointers).
+
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-None
+Should `const impl` blocks additionally generate impls that are not const if any generic
+parameters are not const?
+
+E.g.
+
+```rust
+const impl<T: Add> Add for Foo<T> {
+    fn add(self, other: Self) -> Self {
+        Foo(self.0 + other.0)
+    }
+}
+```
+
+would allow calling `Foo(String::new()) + Foo(String::new())` even though that is (at the time
+of writing this RFC) most definitely not const.
+
+This would go in hand with the current scheme for const functions, which may also be called
+at runtime with runtime arguments, but are checked for soundness as if they were called in
+a const context.
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -1,0 +1,123 @@
+- Feature Name: const_generic_const_fn_bounds
+- Start Date: 2018-10-05
+- RFC PR: (leave this empty)
+- Rust Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Allow `const impl`s for trait impls where all method impls are checked as const fn.
+
+Make it legal to declare trait bounds on generic parameters of `const fn` and allow
+the body of the const fn to call methods on these generic parameters.
+
+# Motivation
+[motivation]: #motivation
+
+Currently one can declare const fns with generic parameters, but one cannot add trait bounds to these
+generic parameters. Thus one is not able to call methods on the generic parameters (or on objects of the
+generic parameter type), because they are fully unconstrained.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+You can call call methods of generic parameters of `const fn`, because they are implicitly assumed to be
+`const fn`. For example, even though the `Add` trait declaration does not contain any mention of `const`,
+you can use it as a trait bound on your generic parameters:
+
+```rust
+const fn triple_add<T: Add>(a: T, b: T, c: T) -> T {
+    a + b + c
+}
+```
+
+The obligation is passed to the caller of your `triple_add` function to supply a type whose `Add` impl is fully
+`const`. Since `Add` only has `add` as a method, in this case one only needs to ensure that the `add` method is
+`const fn`. Instead of adding a `const` modifier to all methods of a trait impl, the modifier is added to the entire
+`impl` block:
+
+```rust
+struct MyInt(i8);
+const impl Add for MyInt {
+    fn add(self, other: Self) -> Self {
+        MyInt(self.0, other.0)
+    }
+}
+```
+
+The const requirement is propagated to all bounds of the impl or its methods,
+so in the following `H` is required to have a const impl of `Hasher`, so that
+methods on `state` are callable.
+
+```rust
+const impl Hash for MyInt {
+    fn hash<H>(
+        &self,
+        state: &mut H,
+    )
+        where H: Hasher
+    {
+        state.write(&[self.0 as u8]);
+    }
+}
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This is the technical portion of the RFC. Explain the design in sufficient detail that:
+
+- Its interaction with other features is clear.
+- It is reasonably clear how the feature would be implemented.
+- Corner cases are dissected by example.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+One cannot add trait bounds to `const fn` without them automatically
+requiring `const impl`s for all monomorphizations. Even if one does not
+call any method on the generic parameter, the methods are still required to be constant.
+
+It is not a fully general design that supports every possible use case,
+but only covers the most common cases. See also the alternatives.
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+## Effect system
+
+A fully powered effect system can allow us to do fine grained constness propagation
+(or no propagation where undesirable). This is way out of scope in the near future
+and this RFC is forward compatible to have its background impl be an effect system.
+
+## Fine grained `const` annotations
+
+One could annotate methods instead of impls, allowing just marking some method impls
+as const fn. This would require some sort of "const bounds" in generic functions that
+can be applied to specific methods. E.g. `where <T as Add>::add: const` or something of
+the sort.
+
+## Explicit `const` bounds
+
+One could require `T: const Trait` bounds to differentiate between bounds on which methods
+can be called and bounds on which no methods can be called. This can backwards compatibly be
+added as an opt-out via `T: ?const Trait` if the desire for such differences is strong.
+
+## Infer all the things
+
+We can just throw all this complexity out the door and allow calling any method on generic
+parameters without an extra annotation `iff` that method satisfies `const fn`. So we'd still
+annotate methods in trait impls, but we would not block calling a function on whether the
+generic parameters fulfill some sort of constness rules. Instead we'd catch this during
+const evaluation.
+
+This is strictly the most powerful and generic variant, but is an enormous backwards compatibility
+hazard as changing a const fn's body to suddenly call a method that it did not before can break
+users of the function.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+None
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -22,7 +22,7 @@ generic parameter type), because they are fully unconstrained.
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-You can call call methods of generic parameters of a const function, because they are implicitly assumed to be
+You can call methods of generic parameters of a const function, because they are implicitly assumed to be
 `const fn`. For example, the `Add` trait declaration has an additional `const` before the trait name, so
 you can use it as a trait bound on your generic parameters:
 
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -70,7 +70,7 @@ for the associated type:
 trait Foo {
     type Bar: Add;
 }
-impl Foo for A {
+impl const Foo for A {
     type Bar = B; // B must have an `impl const Add for B`
 }
 ```
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -117,7 +117,7 @@ struct Bar(Foo);
 impl const Drop for Foo { fn drop(&mut self) {} } // not allowed
 ```
 
-## Runtime uses don't have `const` restrictions?
+## Runtime uses don't have `const` restrictions
 
 `impl const` blocks additionally generate impls that are not const if any generic
 parameters are not const.
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -139,7 +139,7 @@ effect syntax nor effect semantics are proposed by this RFC, the following is ju
 purposes):
 
 ```rust
-impl<constness c, T: const(c) Add> const(c) Add for Foo<T> {
+impl<c: constness, T: const(c) Add> const(c) Add for Foo<T> {
     const(c) fn add(self, other: Self) -> Self {
         Foo(self.0 + other.0)
     }
@@ -166,7 +166,7 @@ const fn add<T: Add>(a: T, b: T) -> T {
 Using the same effect syntax from above:
 
 ```rust
-<constness c> const(c) fn add<T: const(c) Add>(a: T, b: T) -> T {
+<c: constness> const(c) fn add<T: const(c) Add>(a: T, b: T) -> T {
     a + b
 }
 ```
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -153,3 +153,8 @@ of writing this RFC) most definitely not const.
 This would go in hand with the current scheme for const functions, which may also be called
 at runtime with runtime arguments, but are checked for soundness as if they were called in
 a const context.
+
+## Drop
+
+Should we also allow `(SomeDropType, 42).1` as an expression if `SomeDropType`'s `Drop` impl
+was declared with `const impl Drop`?
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -83,6 +83,16 @@ Then you are allowed to actually let a value of `SomeDropType` get dropped withi
 evaluation. This means `(SomeDropType(&Cell::new(42)), 42).1` is now allowed, because we can prove
 that everything from the creation of the value to the destruction is const evaluable.
 
+Note that all fields of types with a `const Drop` impl must have `const Drop` impls, too, as the
+compiler will automatically generate `Drop::drop` calls to the fields:
+
+```rust
+struct Foo;
+impl Drop for Foo { fn drop(&mut self) {} }
+struct Bar(Foo);
+impl const Drop for Foo { fn drop(&mut self) {} } // not allowed
+```
+
 ## Runtime uses don't have `const` restrictions?
 
 `impl const` blocks additionally generate impls that are not const if any generic
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -87,7 +87,9 @@ The implementation of this RFC is (in contrast to some of its alternatives) most
 changes around the syntax of the language (adding `const` modifiers in a few places)
 and ensuring that lowering to HIR and MIR keeps track of that.
 The miri engine already fully supports calling methods on generic
-bounds, there's just no way of declaring them.
+bounds, there's just no way of declaring them. Checking methods for constness is already implemented
+for inherent methods. The implementation will have to extend those checks to also run on methods
+of `const impl` items.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -119,8 +119,8 @@ impl const Drop for Foo { fn drop(&mut self) {} } // not allowed
 
 ## Runtime uses don't have `const` restrictions
 
-`impl const` blocks additionally generate impls that are not const if any generic
-parameters are not const.
+`impl const` blocks are treated as if the constness is a generic parameter
+(see also effect systems in the alternatives).
 
 E.g.
 
@@ -134,7 +134,23 @@ impl<T: Add> const Add for Foo<T> {
 
 allows calling `Foo(String::from("foo")) + Foo(String::from("bar"))` even though that is (at the time
 of writing this RFC) most definitely not const, because `String` only has an `impl Add for String`
-and not an `impl const Add for String`.
+and not an `impl const Add for String`. Expressed in some sort of effect system syntax (neither
+effect syntax nor effect semantics are proposed by this RFC, the following is just for demonstration
+purposes):
+
+```rust
+impl<constness c, T: const(c) Add> const(c) Add for Foo<T> {
+    const(c) fn add(self, other: Self) -> Self {
+        Foo(self.0 + other.0)
+    }
+}
+```
+
+In this scheme on can see that if the `c` parameter is set to `const`, the `T` parameter requires a
+`const Add` bound, and creates a `const Add` impl for `Foo<T>` which then has a `const fn add`
+method. On the other hand, if `c` is `?const`, we get a regular impl without any constness anywhere.
+Of course for regular impls one can still pass a `T` which has a `const Add` impl, but that won't
+cause any constness for `Foo<T>`.
 
 This goes in hand with the current scheme for const functions, which may also be called
 at runtime with runtime arguments, but are checked for soundness as if they were called in
@@ -146,6 +162,18 @@ const fn add<T: Add>(a: T, b: T) -> T {
     a + b
 }
 ```
+
+Using the same effect syntax from above:
+
+```rust
+<constness c> const(c) fn add<T: const(c) Add>(a: T, b: T) -> T {
+    a + b
+}
+```
+
+Here the value of `c` decides both whether the `add` function is `const` and whether its parameter
+`T` has a `const Add` impl. Since both use the same `constness` variable, `T` is guaranteed to have
+a `const Add` iff `add` is `const`.
 
 This feature could have been added in the future in a backwards compatible manner, but without it
 the use of `const` impls is very restricted for the generic types of the standard library due to
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -46,22 +46,25 @@ impl const Add for MyInt {
 }
 ```
 
-The const requirement is propagated to all bounds of the impl or its methods,
+The const requirement is required on all bounds of the impl and its methods,
 so in the following `H` is required to have a const impl of `Hasher`, so that
 methods on `state` are callable.
 
 ```rust
 impl const Hash for MyInt {
-    fn hash<H>(
+    const fn hash<H>(
         &self,
         state: &mut H,
     )
-        where H: Hasher
+        where H: const Hasher
     {
         state.write(&[self.0 as u8]);
     }
 }
 ```
+
+While these `const` keywords could be inferred (after all, they are required), requiring them is
+forward compatible to schemes in the future that allow more fine grained control.
 
 ## Drop
 
@@ -80,16 +83,66 @@ Then you are allowed to actually let a value of `SomeDropType` get dropped withi
 evaluation. This means `(SomeDropType(&Cell::new(42)), 42).1` is now allowed, because we can prove
 that everything from the creation of the value to the destruction is const evaluable.
 
+## Runtime uses don't have `const` restrictions?
+
+`impl const` blocks additionally generate impls that are not const if any generic
+parameters are not const.
+
+E.g.
+
+```rust
+impl<T: const Add> const Add for Foo<T> {
+    fn add(self, other: Self) -> Self {
+        Foo(self.0 + other.0)
+    }
+}
+```
+
+allows calling `Foo(String::from("foo")) + Foo(String::from("bar"))` even though that is (at the time
+of writing this RFC) most definitely not const, because `String` only has an `impl Add for String`
+and not an `impl const Add for String`.
+
+This goes in hand with the current scheme for const functions, which may also be called
+at runtime with runtime arguments, but are checked for soundness as if they were called in
+a const context. E.g. the following function may be called as
+`add(String::from("foo"), String::from("bar"))` at runtime.
+
+```rust
+const fn add<T: const Add>(a: T, b: T) -> T {
+    a + b
+}
+```
+
+This feature could have been added in the future in a backwards compatible manner, but without it
+the use of `const` impls is very restricted for the generic types of the standard library due to
+backwards compatibility.
+Changing an impl to only allow generic types which have a `const` impl for their bounds would break
+situations like the one described above.
+
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
 The implementation of this RFC is (in contrast to some of its alternatives) mostly
-changes around the syntax of the language (adding `const` modifiers in a few places)
+changes around the syntax of the language (allowing `const` modifiers in a few places)
 and ensuring that lowering to HIR and MIR keeps track of that.
 The miri engine already fully supports calling methods on generic
 bounds, there's just no way of declaring them. Checking methods for constness is already implemented
 for inherent methods. The implementation will have to extend those checks to also run on methods
 of `impl const` items.
+
+## Implementation instructions
+
+1. Add an `is_const` field to the AST's `TraitRef`
+2. Adjust the Parser to support `const` modifiers before trait bounds
+3. Add an `is_const` field to the HIR's `TraitRef`
+4. Adjust lowering to pass through the `is_const` field from AST to HIR
+5. Add a a check to `librustc_typeck/check/wfcheck.rs` ensuring that all generic bounds
+    in an `impl const` block have the `in_const` flag set and all methods' `constness` field is
+    `Const`.
+6. Feature gate instead of ban `Predicate::Trait` other than `Sized` in
+    `librustc_mir/transform/qualify_min_const_fn.rs`
+7. Remove the call in https://github.com/rust-lang/rust/blob/f8caa321c7c7214a6c5415e4b3694e65b4ff73a7/src/librustc_passes/ast_validation.rs#L306
+8. Adjust the reference and the book to reflect these changes.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -211,45 +264,4 @@ and `const` modifiers on `impl` blocks.
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-## Runtime uses don't have `const` restrictions?
-
-Should `impl const` blocks additionally generate impls that are not const if any generic
-parameters are not const?
-
-E.g.
-
-```rust
-impl<T: Add> const Add for Foo<T> {
-    fn add(self, other: Self) -> Self {
-        Foo(self.0 + other.0)
-    }
-}
-```
-
-would allow calling `Foo(String::from("foo")) + Foo(String::from("bar"))` even though that is (at the time
-of writing this RFC) most definitely not const, because `String` only has an `impl Add for String`
-and not an `impl const Add for String`.
-
-This would go in hand with the current scheme for const functions, which may also be called
-at runtime with runtime arguments, but are checked for soundness as if they were called in
-a const context.
-
-## Require `const` bounds on everything inside an `impl const` block?
-
-Instead of inferring `const`ness on all bounds and functions inside a `impl const` block,
-we force the user to supply these bounds. This is more consistent with not inferring `const`
-on `const` function argument types and generic bounds. The `Hash` example from above would
-then look like
-
-```rust
-impl const Hash for MyInt {
-    const fn hash<H>(
-        &self,
-        state: &mut H,
-    )
-        where H: const Hasher
-    {
-        state.write(&[self.0 as u8]);
-    }
-}
-```
+Everything has been addressed in the reviews
--- a/const-generic-const-fn-bounds.md
+++ b/const-generic-const-fn-bounds.md
@@ -6,7 +6,7 @@
 # Summary
 [summary]: #summary
 
-Allow `const impl`s for trait impls where all method impls are checked as const fn.
+Allow `impl const Trait` for trait impls where all method impls are checked as const fn.
 
 Make it legal to declare trait bounds on generic parameters of const functions and allow
 the body of the const fn to call methods on the generic parameters that have a `const` modifier
@@ -39,7 +39,7 @@ The obligation is passed to the caller of your `triple_add` function to supply a
 
 ```rust
 struct MyInt(i8);
-const impl Add for MyInt {
+impl const Add for MyInt {
     fn add(self, other: Self) -> Self {
         MyInt(self.0 + other.0)
     }
@@ -51,7 +51,7 @@ so in the following `H` is required to have a const impl of `Hasher`, so that
 methods on `state` are callable.
 
 ```rust
-const impl Hash for MyInt {
+impl const Hash for MyInt {
     fn hash<H>(
         &self,
         state: &mut H,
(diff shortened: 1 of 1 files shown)


gnzlbg:
From all options proposed, this appears to be the most minimal one that gets the problem solved, and it also appears to be forward compatible with all others (supporting traits with const methods, an effect system, etc.). 

Its a bit of a shame that all methods of an implementation must be const, but one has to start somewhere, and even if we were to relax this in the future, we might still want "sugar" for the cases in which the user wants to implement all methods as const anyways. 

So thumbs up from me.

gnzlbg:
Would it make sense to make it obvious in the bound that the impl must be `const` ? For example using `T: const impl Foo` ? We might want to allow requiring `const` impls in non-const functions in the future:

```rust
trait Bar { fn bar(&self) -> usize; }
/*not const*/ fn foo<T: const impl Bar>(x: T) {
      const X: usize = x.bar();
}
```

oli-obk:
Ah I have no thought about this use-case. It probably is more sensible to stay forward compatible by using the explicit annotation alternative instead of punting to an opt out in the future. I'll adjust the RFC accordingly

alexreg:
Very glad to see an RFC for this finally! Looks like a good path forward to me. Effect systems (for const and other things) would be great in the future, but yes, that involves significant technical investment... this is a good compromise for now.

RalfJung:
I like most of it, but I think I'd prefer this alternative:

> One could require no const on the bounds (e.g. T: Trait) and assume constness for all bounds. An opt-out via T: ?const Trait would then allow declaring bounds that cannot be used for calling methods. This design causes discrepancies with const fn pointers as arguments (where the constness would be needed, as normal function pointers already exist as the type of constants). Also it is not forward compatible to allowing const trait bounds on non-const functions

The thing is, you already kind-of have all the complexity of this alternative in your proposal when you say

> The const requirement is propagated to all bounds of the impl or its methods,

I find it rather inconsistent to do automatic propagation on `const` bounds on `const impl` but not on `const fn`.

Moreover, I assume that your `triple_add` is callable at run-time even with a non-`const` `impl`, right? So the bound is also somewhat misleading.

oli-obk:
> I find it rather inconsistent to do automatic propagation on const bounds on const impl but not on const fn.

True. We could require manually annotating everything inside `const impl`s for now to stay consistent and forward compatible.

eddyb:
I don't think this provides a workable solution for deriving (short of `#[derive(const Clone)]` - which I don't hate, but I don't know if it fits in with everything - cc @petrochenkov @alexcrichton).

But also, I think it should be stated more clearly what the difference is between `const fn` callers versus runtime `fn` callers, and how supertraits and implied bounds interact with this proposal.

eddyb:
Another drawback to keep in mind is the syntactic similarity to `const X: T` params.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,160 @@
...
+of writing this RFC) most definitely not const.
+
+This would go in hand with the current scheme for const functions, which may also be called
+at runtime with runtime arguments, but are checked for soundness as if they were called in
+a const context.
+
+## Drop
+
+Should we also allow `(SomeDropType, 42).1` as an expression if `SomeDropType`'s `Drop` impl
+was declared with `const impl Drop`?
```
eddyb: Yes! That's literally what it's been waiting for :tada:

oli-obk:
> I don't think this provides a workable solution for deriving (short of #[derive(const Clone)] - which I don't hate, but I don't know if it fits in with everything

There's no way around this. Imagine the following situation:

```rust
struct Foo;
const impl Clone for Foo {
    fn clone(&self) -> Self { Foo }
}
#[derive(Clone)]
pub struct Bar(Foo);
```

If we automagically made the derived `Clone` impl for `Bar` const, and internally changed the `const impl Clone for Foo` to `impl Clone for Foo`, we'd be breaking users of the crate that used `Bar::clone` in const contexts. This is exactly why we have `const fn` instead of inferring the constness from the body: semver compatibility.

const-generic-const-fn-bounds.md
```diff
@@ -131,6 +131,46 @@ It might also be desirable to make the automatic `Fn*` impls on function types a
...
+const impl Clone for Bar {
+    fn clone(&self) -> Self { Bar }
+}
+```
+
+could theoretically have a scheme inferring `Foo`'s `Clone` impl to be `const`. If some time
+later the `const impl Clone for Bar` (a private type) is changed to just `impl`, `Foo`'s `Clone`
+impl would suddenly stop being `const`, without any visible change to the API. This should not
+be allowed for the same reason as why we're not inferring `const` on functions: changes to private
+things should not affect the constness of public things, because that is not compatible with semver.
```
eddyb: Hmm, is the need for something like `#[derive(const Clone)]` mentioned anywhere explicitly?
I mean, this section looks like the motivation, but not the conclusion.

eddyb: Right, I just don't see `derive(const Foo)` mentioned at all.

oli-obk:
@centril I think this is ready for opening a full RFC in the RFC repo. Do you want to do a review before I do this?

oli-obk:
@scottmcm brought up a point on discord:

We have `unsafe impl Foo for Bar {}`, which means that it's an impl for an `unsafe trait Foo`, not that all methods being implemented are `unsafe`.

In contrast to that, `unsafe fn` means the function can only be called in unsafe blocks and `const fn` means the function can be called in const contexts.

So `const impl Foo for Bar {}` gives a new meaning to a modifier in a place where another modifier has a different meaning.

I am changing this RFC to use `impl const Foo for Bar {}`, which has the nice symmetry of having the `const` infront of the trait name like in bounds (`T: const Trait`).

scottmcm:
This looks pretty good; the one additional thing I'd like to see is a plan for how to write generic impls  when a generic type has a impl for a const trait.  For example, how do we have `std::cmp::rev` impl `Ord` or `const Ord` appropriately and without breaking anyone?

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,255 @@
...
+## Specialization
+
+Impl specialization is still unstable. There should be a separate RFC for declaring how
+const impl blocks and specialization interact. For now one may not have both `default`
+and `const` modifiers on `impl` blocks.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+## Runtime uses don't have `const` restrictions?
```
oli-obk: > how do we have std::cmp::Rev impl Ord or const Ord appropriately and without breaking anyone?

This is discussed in this section. You'd write the impl as follows:

```rust
impl<T: Ord> const Ord for Reverse<T> {
    #[inline]
    fn cmp(&self, other: &Reverse<T>) -> Ordering {
        other.0.cmp(&self.0)
    }
}
```

compare this to the current impl in https://doc.rust-lang.org/src/core/cmp.rs.html#458-463

```rust
impl<T: Ord> Ord for Reverse<T> {
    #[inline]
    fn cmp(&self, other: &Reverse<T>) -> Ordering {
        other.0.cmp(&self.0)
    }
}
```

The only difference is the `const` in front of `Ord`. If the unresolved question about requiring `const` bounds becomes canonical we'd have

```rust
impl<T: const Ord> const Ord for Reverse<T> {
    #[inline]
    fn cmp(&self, other: &Reverse<T>) -> Ordering {
        other.0.cmp(&self.0)
    }
}
```

If you pass a `T` that is not const, you also don't get a `const` impl, just like calling a `const` function outside a const context gives you a regular function.

Pratyush:
In this design, how would I go about annotating specific functions within a trait as `const`? Is there some way to do the following?
```rust
trait Foo {
    const fn do_stuff(a: usize) -> usize;
}

const fn bar<T: Foo>(a: usize) -> usize {
    T::do_stuff(a)
}
```

oli-obk:
No, that is orthogonal to this RFC. Marking a function as const fn in a trait means that all impls of the trait will need to provide a const fn. While interesting, I have yet to see a case where this would be useful. This RFC does allow you to define

```rust
trait Bar {
    fn do_stuff(a: usize) -> usize;
}

const fn do_stuff<T: const Bar>(a: usize) -> usize {
    Bar::do_stuff(a)
}
```

which amounts to the same thing, but it's not an associated constant function.

Pratyush:
Thanks for the clarification. The motivation for this is a scenario where you want some items in a trait to be `const`, but not necessarily others, and all reasonable `impl`s of the trait would be able to satisfy `const`-ness for the relevant items. I have some code that could benefit from this sort of fine-grained `const`-ness.

The RFC does actually enable something like this, but in a slightly hacky way:
```rust
trait _Foo {
    fn do_stuff(a: usize) -> usize;
}

trait Foo: const _Foo {
    fn other_fn();
}

const fn bar<T: Foo>(a: usize) -> usize {
    T::do_stuff(a)
}
```

oli-obk:
> Would the above work? Note the lack of a const bound in T: Foo.

That would not really work with just this RFC, since const bounds on trait declarations are not something discussed at all. You can still get that split by only adding a `const` bound for the trait where you actually need the bounds:

```rust
trait FooA {
    fn do_stuff(a: usize) -> usize;
}

trait FooB: FooA {
    fn other_fn();
}

const fn bar<T: FooB + const FooA>(a: usize) -> usize {
    T::do_stuff(a)
}
```

varkor:
I have yet to thoroughly read through the RFC/discussions, but from a naïve intuitive perspective, `impl const Trait` doesn't seem like the right abstraction when we think about const type theoretically. I'll just jot down some notes and try to elaborate afterwards.

First, we have an existing basic notion of types in Rust:
- We have a universe/set of types (called `Type` or `*` or whatever) that we can roughly split into "function types" and "non-function types" (that is, `Type` is a category).
- Traits are type-level functions: they take a type (`Self`) and return a type, which is the signature of the trait with `Self` specialised to the input type.
- Implementations of traits are values of traits applied to types. E.g. `impl Trait for Foo` defines a value, `<Foo as Trait`>, which has the type `Trait(Foo)`. (I'm overloading notation here a little, but hopefully it's clear what I mean.)

Now we want to consider `const` (in its various forms). What do we want from a model of `const`?
- There are strictly more `fn`s than `const fn`s (because every `const fn` can be used as a normal function, but there are some functions inexpressible as `const fn`). So the set of `const fn`s must embed into the set of `fn`s.
- We must be able to convert a `const fn` to a `fn` (as pointed out above). (It doesn't matter here whether that's implicit or explicit.)

It makes sense then to consider a new universe/category of types, `const Type`, which is analogous to `Type`, but with const types (whatever they are) and const functions. This is a smaller universe than `Type`. We have an embedding map (strictly speaking, a functor) taking const types to their "runtime" equivalents and const functions to their runtime equivalents.

This embedding is strictly one-way. We can't go from a runtime type to a const type. Built-in types like `u8`, `()`, `bool`, etc. are const types, and we can use them in `const fn` and normal `fn`s. An example of something that _isn't_ a const type would be a type making use of something we know _isn't_ const: e.g. a pointer to a non-const function.

This essentially means that const types are those that are made up solely of const types, whereas "runtime" types are those made up of const types or other runtime types (the base case for these definitions are in the functions).

(It's critically important that non-const types can't make their way into const types, because otherwise we would introduce type dependency into the type system.)

The crux is this: with this model, const traits are (const) type-level functions on the universe of const types, `const Type`. They take a const type and return a const type. Implementations of const traits are *values* of applications of const traits.

Therefore, const types, const functions and const traits all make sense as notions to consider, because the corresponding non-const notion is well-defined. But a "const impl" only makes sense if you consider it as "an implementation of a const trait". The *implementation* itself can't be const: it's just the definition of a value.

This also applies to the bounds of a type: `A: const Trait` doesn't make sense, because traits are inherently either const or not-const: the constness is part of the `Trait` definition itself. It doesn't make sense to switch between them.

In summary, I don't think it makes sense to have any notion of "const impl" without first having concrete notions of "const trait". (Aside: I'm fairly sure the constness of everything but functions can be inferred without issue in a model like the one I describe, as a side point.)

The RFC as it stands has very little in the way of theoretical motivation and (even if I've misread the RFC entirely and this does all work out) I think that's an important consideration. `const` can be quite subtle, and we want to make sure we get this design right.

oli-obk:
This RFC is the culmination of all the discussion in the const fn tracking issue which made it very clear that having a separate list of const traits and nonconst traits would end up duplicating basically every single trait except for a select few like Rand or io related ones. What this RFC does is not make an impl const, but it makes the type usable in const contexts with respect to all the trait interfaces that have a const impl. 

I agree that from a type theoretical stand point this is neither elegant nor perfect, but it is the only practical solution that I have been able to find that covers most use cases without going fully into an effect system.

One very important point about this RFC is that it is intuitive. You want a const fn to take a generic argument? You can just give it that. You want to also have trait bounds? All good. You want to call methods on the generic argument? Make the trait bounds you want to use `const`. You want to be able to use your type as an argument to such a const fn? Throw a `const` on the relevant trait impls. You want your type to also have other trait impls, but those need nonconst things? All fine.

For example: There is no const fn world out there where we'll ever forbid `usize` in const contexts. There's also already implementations of Rand traits or similar guaranteed nonconst operations for `usize`, and we don't want to forbid those either. There are also a lot of traits which, when implemented for some types are const, and, when implemented for others, are nonconst. We don't want two different traits for these two possibilities.

varkor:
> This RFC is the culmination of all the discussion in the const fn tracking issue which made it very clear that having a separate list of const traits and nonconst traits would end up duplicating basically every single trait except for a select few like Rand or io related ones. What this RFC does is not make an impl const, but it makes the type usable in const contexts with respect to all the trait interfaces that have a const impl.

Thanks, that's helpful to have the context for (there's so much to catch up on with the `const fn` discussions).

> I agree that from a type theoretical stand point this is neither elegant nor perfect, but it is the only practical solution that I have been able to find that covers most use cases without going fully into an effect system.

I think my argument was slightly simplifying things that might make the situation a little nicer theoretically, so I'm going to give it some thought and report back.

> There are also a lot of traits which, when implemented for some types are const, and, when implemented for others, are nonconst. We don't want two different traits for these two possibilities.

That's true. That would not be a pleasant state of affairs. I'll think some more about it.

varkor:
I've sketched up a (very rough) draft of some theoretical grounding for `const` here: https://hackmd.io/s/rJrvLLfzV#const-types (the bit before the `const` heading can be ignored). It's rushed, because I want to check my reasoning before the RFC is actually posted.

In summary, as I see it (I might be overlooking something that has already come up in discussion, so apologies if I'm repeating things):
We can define a notion of `const trait` such that:
- if we implement it for const types (which includes most types — anything that doesn't transitively reference a non-const fn), `const fn`s as the methods, then it can be used as bounds in any `const fn` and also in any `fn`;
- if we implement it for a non-const type, *or* with (non-const) `fn`s as methods, then it can be used as bounds in any (non-const) `fn`.
Most traits will be const. Non-`const` traits are those referencing non-const types and can be implemented for const types or non-const types, but can only be used in bounds for non-const `fn`s.

This way, most traits will be defined once (as `const`). There will be a mix of const and non-const implementations for const traits. (But everything works as expected.)

A natural syntax also suggests itself in this case: `const trait Trait` and `const impl Trait for Type` (where `Trait` is const here). Notably in relation to this RFC, declaring `const` on bounds is entirely unnecessary — it's completely inferrable, unambiguously.

This is a different approach to that in this RFC, but I think it has the same advantages, while having nice theoretic properties too. (If my explanations haven't been clear, it might be easier to discuss synchronously.)

Sorry if I'm completely overlooking something: I don't want to waste time by reiterating discussions! I just think that maybe there's a slightly more natural approach here that it's worth looking into if it hasn't already been.

oli-obk:
Citing your draft:

>  Intuitively, const Type consists of all the types whose values can be computed at compile-time.

We already allow `const FOO: fn() = main;` since stable Rust 1.0. We can neither go back on this nor do I think it's desirable. Or am I misinterpreting your statement?

varkor:
> If I understand it correctly, you want to forbid the following `new` method from being const fn.

> We can neither go back on this nor do I think it's desirable. Or am I misinterpreting your statement?

This is a very good point and you're right, I hadn't considered it.

I think there's a straightforward extension to the model that handles this. We define reference and pointer types (including function pointers) to be references to *any* type (const or not). Reference and pointer types are still either const or non-const types, but their referee types are not necessarily.

So we can have `fn` pointers in const types and functions. However, `fn` pointers take values of non-const types as arguments, which can't exist inside a `const fn`, which explains why we're not able to call `fn` from a const function.

This way, we can always pass around non-const types (strictly speaking, references to them), but we can't "use" them for anything. I think this matches up with what we want to be possible for const types and traits?

oli-obk:
So. In order to fully understand your proposal, I'm translating all examples from this RFC to your proposal as I understand it.

```rust
// the `T` is const, not the `Add`
const fn triple_add<const T: Add>(a: T, b: T, c: T) -> T {
    a + b + c
}
```

We need `const` in definition, otherwise changing private fields can change the constness, which is not semver compatible.

Additionally I'm not sure how to handle having some trait impls which are const and some which are not.

```rust
const struct MyInt(i8);
impl Add for MyInt {
    fn add(self, other: Self) -> Self {
        MyInt(self.0 + other.0)
    }
}
```

While this works, `Add` can't be `const Add`, because we already have `Add` impls for e.g. `String`:

```rust
const struct MyString(*mut u8, usize);
impl MyString {
    fn new() -> Self {
        MyString(std::ptr::null_mut(), 0)
    }
    fn from(s: &str) -> Self {
        // nonconst operation
        MyString(s.bytes().to_vec().into_boxed_slice().into_raw(), s.len())
    }
}
impl Add for MyString {
    fn add(self, other: Self) -> Self {
        // do some allocation code and create a new string. also nonconst op
    }
}
```

How do we distinguish between these different add impls?

We already have a lot of types which have some const fns and some nonconst fns. How does this mix fit into the notion of "const types"?

varkor:
> We need `const` in the definition, otherwise changing private fields can change the constness, which is not semver compatible.

Yes, that seems reasonable.

> Additionally I'm not sure how to handle having some trait impls which are const and some which are not.

I think my proposal would be something like (the syntax is flexible):
```rust
const struct MyInt(i8);

// `Add` is a const trait, whose methods are all `const fn`
const trait Add {
	fn add(self, other: Self) -> Self;
}

// We're implementing a const trait for a const type and all
// the methods are const-safe, so this is a const impl.
const impl Add for MyInt {
    fn add(self, other: Self) -> Self {
        MyInt(self.0 + other.0)
    }
}
```

Now we want to implement `Add` for `String`. I didn't mention non-const inherent impls for const types, but it should work identically to non-const trait impls for const types.
```rust
const struct MyString(*mut u8, usize);

impl MyString {
    // As in your example, with non-const functions.
}

// `MyString` is const and `Add` is const, but this impl is not going to be,
// because its methods are not all const.
impl Add for MyString {
    fn add(self, other: Self) -> Self {
        // As before.
    }
}
```

Now the trait bound `MyInt: Add` holds in const bounds and in non-const bounds. However, the trait bound `MyString: Add` only holds in non-const bounds (despite `Add` and `MyString` separately being const). I think although we need `const` in the definition of functions, types, traits and impls for the semver reasons you mention, we shouldn't need it in bounds. (The only way it can change is if any one of the type, trait or impl changes its constness, which is already a semver breaking change.)

That is, `A: Trait` always requires `A`, `Trait` and the impl of `Trait` for `A` to be const, if it's used in a const parameter bound; otherwise it's a non-const bound.

> We already have a lot of types which have some const fns and some nonconst fns. How does this mix fit into the notion of "const types"?

This is the part I only lightly touched upon in the notes, so I'd want to go over the ideas to make sure they're sound, but I think to model it we have a slightly more complex universe of types, where non-const types can be a product of types and const types and the projections (i.e. property accesses / method calls) have the correct constness. So anything with *any* non-const part is automatically non-const, but taking const subparts of them preserves constness.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,255 @@
...
+[motivation]: #motivation
+
+Currently one can declare const fns with generic parameters, but one cannot add trait bounds to these
+generic parameters. Thus one is not able to call methods on the generic parameters (or on objects of the
+generic parameter type), because they are fully unconstrained.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+You can call call methods of generic parameters of a const function, because they are implicitly assumed to be
```
varkor:
```suggestion
You can call methods of generic parameters of a const function, because they are implicitly assumed to be
```

varkor:
> What is the difference between your proposal and this RFC execept some syntactical [details]

Apart from the concept of "const types" and "const traits", I think my proposed model does almost coincide with this RFC now 🙂

The question is whether, syntactically, `const type` and `const trait` are useful distinctions to make, or whether all user-defined types and traits are automatically const. (E.g. is there any reason it'd be useful to make a trait explicitly _not_ const?) If they are, then I don't think there are non-syntactic differences between this RFC and my model.

> (I don't see how in the above definition there would be any types which we won't add const to)

Yes, I think that's the case too. I think it's also probably the case for traits too (even with defaults).

---

It may be that we've arrived back at the original design, in which case thanks for humouring me. I personally feel better about the design now that I think it _does_ have a solid backing in the type system — and this could be something to add to the **Rationale** section. (I think it provides more motivation for the design than the alternatives, e.g. why effect systems are not necessary and exactly what needs to be marked `const`; and also gives another way of looking at what the "natural solutions" are to the future work / unresolved questions.)

I do think in light of this perspective, a slightly different syntax does make more sense, but being on the same page about the strength of the design in general is nice.

varkor:
As a follow up, I think the ideal syntax is:
- `const impl T { .. }`, `const impl Trait for T { .. }`
- No `const` prefixing types or traits (anywhere, either in definitions or in parameters and bounds), as all types and traits are implicitly `const`.
- No `const` in bounds, as the constness is directly (unambiguously) inferable from the constness of the function.

In addition, `const impl` *is* consistent with `unsafe impl` if you view `unsafe impl` as:
- It is unsafe to implement this unsafe trait (rather than "we're implementing an unsafe trait").
So there's no issue with consistency there.

rpjohnst:
So if this syntax...

```rust
impl const Hash for MyInt {
    const fn hash<H>(&self, state: &mut H) where
        H: const Hasher
    {
        state.write(&[self.0 as u8]);
    }
}
```

...means that you can call `MyInt::hash` with a non-`const` `H` at runtime, would it ever be possible to express "you can only call `MyInt::hash` with a `const` `H`, even at runtime"? This seems related to, maybe kind of an inverse of, the `T: ?const Trait` described in the alternatives section.

This would be expressible in a hypothetical effect-based implementation, by making `hash` polymorphically `const` (as the proposal does with the syntax above) while instead fixing `H: const Hasher` as `const`. However, I would expect such an implementation to use the `H: const Hasher` syntax to do so.

One more forward-compatible syntax would be the one from the "No explicit `const` bounds" alternative, though I admit I don't understand either of the problems listed there. This way instead of a hidden `const` bound, there could be a hidden effect inference variable, which could be explicitly replaced with *either* `const` (forcing it to be `const` even at runtime, as above) *or* `?const` (forcing it to allow non-`const` impls, even at const-eval time).

alexreg:
@varkor Is this still backwards-compatible with adding an effect system later?

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,279 @@
...
+as const fn. This would require some sort of "const bounds" in generic functions that
+can be applied to specific methods. E.g. `where <T as Add>::add: const` or something of
+the sort. This design is more complex than the current one and we'd probably want the
+current one as sugar anyway
+
+## No explicit `const` bounds
+
+One could require no `const` on the bounds (e.g. `T: Trait`) and assume constness for all
+bounds. An opt-out via `T: ?const Trait` would then allow declaring bounds that cannot be
+used for calling methods. This design causes discrepancies with `const fn` pointers as
```
oli-obk: The discrepancy is that 

```rust
impl const Trait for Type {
    fn foo<T: OtherTrait>(t: T) { ... }
}
```

would require `OtherTrait` to be `const` automatically, while

```rust
const fn foo<T: OtherTrait>(t: T) { ... }
```

would not do that. That is very inconsistent and will lead to a lot of confusion.

For semver purposes (which is the only reason we even have the `const` keyword for function and are considering it for bounds and impls) we either need all bounds on methods of an `impl const Trait` to be `const`, too, or we need none of them to be `const`, which would be very sad because it would make `impl const Hash` impossible to implement (other than outright panicking or making it do nothing, which is wrong).

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,279 @@
...
+arguments (where the constness would be needed, as normal function pointers already exist
+as the type of constants). Also it is not forward compatible to allowing `const` trait bounds
```
oli-obk: So the open question from the last few comments on this RFC is "can we do the following?"

```rust
fn foo<T: const Trait>() -> u32 {
    const FOO: u32 = T::some_method();
    FOO
}
```

Which is totally out of scope of this RFC, because we can't even do

```rust
fn foo<T: Trait>() -> u32 {
    const FOO: u32 = T::SomeAssocConst;
    FOO
}
```

(see [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cf4d2d436dc64fc274cc0c7e013ee1a8))

```
error[E0401]: can't use type parameters from outer function
 --> src/lib.rs:6:22
  |
5 | fn foo<T: Trait>() -> u32 {
  |    --- - type variable from outer function
  |    |
  |    try adding a local type parameter in this method instead
6 |     const FOO: u32 = T::SomeAssocConst;
  |                      ^^^^^^^^^^^^^^^^^ use of type variable from outer function
```

This boils down to the same question as whether we want to be able to support

```rust
fn foo(f: const fn() -> u32) {
    const FOO: u32 = f();
}
```

Which goes into the territory of generic value arguments and should be solved there instead of here.

cramertj:
Sorry to bring this back up, but I only just discovered the existence of this PR :)

@oli-obk 
> So const impl Foo for Bar {} gives a new meaning to a modifier in a place where another modifier has a different meaning.

> I am changing this RFC to use impl const Foo for Bar {}, which has the nice symmetry of having the const infront of the trait name like in bounds (T: const Trait).

Note that we also have `default impl Foo for Bar { ... }`-- IMO we should be consistent in putting the modifiers before `impl`, not after.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,279 @@
...
+```rust
+struct SomeDropType<'a>(&'a Cell<u32>);
+impl const Drop for SomeDropType {
+    fn drop(&mut self) {
+        self.0.set(self.0.get() - 1);
+    }
+}
+```
+
+Then you are allowed to actually let a value of `SomeDropType` get dropped within a constant
```
cramertj: Does the `const Drop` impl also require then that `SomeDropType`'s fields are all `const`-droppable as well?

cramertj: (in contrast to just checking that `fn drop` is const-compatible)

const-generic-const-fn-bounds.md
rpjohnst: Trait bounds on `const fn`s aren't stabilized yet, so doesn't this just mean that we should apply the same rules there, and have `const fn foo<T: OtherTrait>(t: T)` require `OtherTrait` to be `const` unless called at runtime?

I am not sure what the last part of your comment means. `impl const Hash for T` does need `hash`'s `H: Hasher` to be `const`, sure, but that doesn't mean *every* trait will *always* be like that, or that *every* trait will *always* want this "`const` polymorphism" to remove that requirement at runtime.

varkor:
After some additional discussion, I've given this quite a lot more thought and decided to write up what I think is the ideal design for this problem: [const types, traits and implementations in Rust](https://varkor.github.io/blog/2019/01/11/const-types-traits-and-implementations-in-Rust.html).

It has many similarities to the current proposal here, but also differs in some significant ways: for example, it is more expressive in what functions may be called at run-time.

I'd especially love to hear your thoughts, @oli-obk!

varkor:
> Every individual point in the RFC and that blog post makes sense, but it wasn't clear to me where the blog post was trying to agree with the RFC as opposed to disagree with it as opposed to describe an unproposed alternative only to reject it.

Ah, you're completely right @Ixrec — this is something I should have clarified.

> It seems like the blog post only proposes two concrete changes to syntax/semantics of the RFC:

I think there are four major changes proposed by the post, with respect to this RFC. The two you said, but also the addition of `const trait` and const trait methods:
- `const impl` rather than `impl const`.
- Explicitly `const` bounds.
- `const` trait methods for individually required-const methods.
- `const trait` as syntactic sugar.

However, the point about when `const` should be required in bounds is worth clarifying. The proposal in the post is supposed to require `const` precisely when necessary (not too many `const`s, nor too few). Conversely, the RFC requires more `const`s than necessary. However, the approach proposed in the RFC is not forwards compatible with some cases. It states:

> impl const blocks additionally generate impls that are not const if any generic parameters are not const.

This fails the examples in the post where a constant definition is made inside a run-time function, which still requires the const bound even after being called at run-time. (The generic parameters are not const in that case.)

You need the level of expressiveness where you can decide which bounds remain const at run-time. (Alternatively, if you require const function parameter bounds to be const at run-time too, you're unable to express some more complex const implementations, like that for `Hash`.)

Therefore, I argue that not getting the number of `const` declarations right from the beginning is going to lead to forwards-incompatibility (or, in a best-case scenario, simply severely limit what implementations you can define).

---

In summary, I think you covered the main points well! It's just worth pointing out that a conservative design is harder to get right than you'd hope, and I think this RFC falls afoul of that right now.

oli-obk:
Your blog post solves various issues with this RFC in a clean way and I now have come around to seeing that `?const` is not just preferrable, but the only way to implement all the use cases.

The only issue I see with the blog post is the fact that "if all methods in an impl are `const`, you can use that type where a `const Trait` is required". E.g.

```rust
const impl V for P {
    fn foo(C) -> D;
    fn bar(E) -> F;
}
// ...desugars to...
impl V for P {
    const fn foo(C) -> D;
    const fn bar(E) -> F;
}
```

breaks the nice feature that adding a new method to a trait is not a breaking change as long as it has a default impl. With your scheme of inferring constness depending on whether all methods are `const`, you'd be requiring any new methods on traits to have a `const` default impl, which is clearly a step down from the status quo. Thus I believe we should stay with the `const impl` or `impl const` scheme.

Which brings us to the bikeshed painting. It's not great that we could end up with

```rust
default const unsafe impl Foo for Bar {}
```

But that's going to be the very very rare case, so it's not really an issue. Mostly we're going to have

```rust
const impl Foo for Bar {}
```

and while the RFC uses `impl const Foo for Bar` for consistency with `T: const Foo` bounds, users will likely try `const impl` due to `const fn` muscle memory. I don't have a strong opinion either way.

const-generic-const-fn-bounds.md
oli-obk: > Why does this relate to generic value arguments

The thing is that generic arguments of functions do not influence any items inside that function, unlike generic arguments of `impl` blocks, which allow using their generic arguments on their associated items. Regular function arguments (or const function arguments) being used in inner items boils down to the same thing (semantically) while being something utterly different syntactically and intuitively.

> why does that make this incompatible with const trait bounds on non-const functions?

It doesn't make it incompatible, these bounds just don't make any sense in the way the language is set up so far. I think an example demonstrating the problem more closely is

```rust
fn foo<T: Trait>() -> [u8; 3] {
    [42; T::SomeAssocConst]
}
```

This code will only compile if given a `T` argument whose `SomeAssocConst` constant is `3`. Such behavior is neither desirable (very duck-typy, because the bounds on the function don't show that this behavior is expected), nor easy to implement (is `T::FOO + 1` the same thing as `1 + T::Foo`? Or `2 * T::FOO` the same thing as `T::FOO + T::FOO`?)

const-generic-const-fn-bounds.md
oli-obk: > every trait will always be like that, or that every trait will always want this "const polymorphism" to remove that requirement at runtime.

Yes, I see the problems with my approach now. Thanks

const-generic-const-fn-bounds.md
oli-obk: Yes, I overlooked that. They must be, as drop glue would be invoking their `Drop` impls, too.

Ixrec:
> Your blog post solves various issues with this RFC in a clean way and I now have come around to seeing that ?const is not just preferrable, but the only way to implement all the use cases.

I'm confused. I thought the tweaks proposed by the blog post eliminated any potential use case for `?const` syntax, and that was one of its benefits.

varkor:
> With your scheme of inferring constness depending on whether all methods are const, you'd be requiring any new methods on traits to have a const default impl, which is clearly a step down from the status quo. Thus I believe we should stay with the const impl or impl const scheme.

Yes, I think this is probably the better option too. I briefly considered this under "API considerations", but didn't commit to it — but I think you're right in that requiring `const impl`/`impl const` is better behaviour.

> I'm confused. I thought the tweaks proposed by the blog post eliminated any potential use case for `?const syntax`, and that was one of its benefits.

I updated the post, because it was pointed out on Reddit that occasionally forcing a `const` bound on a `const impl` was too strict and it would be sufficient to simply make use of items declared `const` in the trait. It was nice only having a single modifier, but `?const` mirrors `?Sized` so it doesn't feel unintuitive (and I think it probably won't need to be used frequently, apart from by API authors).

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,398 @@
...
+
+```rust
+struct Foo<T: Trait>(T);
+const fn new<T: Trait>(t: T) -> Foo<T> {
+    Foo(t)
+}
+```
+
+Unfortunately, with the given syntax in this RFC, one can now only call the `new` method if `T` has
+an `impl const Trait for T { ... }`. Thus an opt-out similar to `?Sized` can be used:
```
rpjohnst: This confused me for a second, might be good to rephrase so it's obvious you mean you can only call the `new` method *at `const` time* if `T: const Trait`.

const-generic-const-fn-bounds.md
cramertj: Presumably types which don't implement `Drop` manually but have fields that do automatically inherit `const`-`Drop`-ness from their fields? How would one specify a `const`-`Drop`-glue bound without requiring types to implement the actual `Drop` trait?

oli-obk: hm. good question. that is an issue. We can't infer it, because that would break code when you change a private field from `const Drop` to just `Drop`. I do not have a good idea for this situation. Suggestions are very welcome.

cramertj: It's gross, but maybe `const drop` for things that have const-compatible `drop` glue rather than `const Drop` for things that `const`-ly implement the `Drop` trait? That's confusing as heck, sorry, but I don't have other good ideas.

const-generic-const-fn-bounds.md
oli-obk: done

oli-obk:
I don't see why 

> const traits.

are necessary for a general design. Can you elaborate? I extended the future work block for this topic, so maybe comment directly on it.

>  it seems unnatural not to immediately go for the most general design (though I appreciate this proposal is forwards compatible as far as I can tell with the general design).

The idea is that we go with the design that covers most use cases. I don't know if we really need the full complexity of all extensions mentioned in the future work section. I do like minimal approaches, but I'm not against creating an E-RFC for the full deal.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,496 @@
...
+
+```rust
+const fn foo<T: ?const Foo>() -> i32 {
+    T::a()
+}
+```
+
+even though the `?const` modifier explicitly opts out of constness.
+
+The author of this RFC believes this feature to be unnecessary, since one can get the same effect
```
varkor: I agree. The `const` in traits is nice sugar as it allows all semantically-related parts of the interface to be included in a single trait, but it is just sugar.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,496 @@
...
+trait V {
+    const fn foo(C) -> D;
+    const fn bar(E) -> F;
+}
+```
+
+## `?const` modifiers in trait methods
+
+This RFC does not touch `trait` methods at all, all traits are defined as they would be defined
+without `const` functions existing. A future extension could allow
```
varkor: Agree that this is only necessary/warranted if/when trait methods can also be declared `const`.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,496 @@
...
+replace `T`. This allows `const` trait bounds on any (even non-const) functions, e.g. in
+
+```rust
+fn foo<T: const Bar>() -> i32 {
+    const FOO: i32 = T::bar();
+    FOO
+}
+```
+
+Which, once `const` items and array lengths inside of functions can make use of the generics of
```
varkor: Is there a reason generics cannot be used inside `const` contexts?

Also, even without directly accessing the parameters, we have things like:
```rust
fn foo<T: const Bar>(t: T) -> i32 {
    const FOO: i32 = t.baz();
    FOO
}
```
This seems like an addition that goes hand-in-hand with the rest of the RFC.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,496 @@
...
+method default body as being `const`. The author of this RFC considers prepending the default body's
+method signature with `const` to be the most intuitive syntax.
+
+```rust
+trait Foo {
+    const fn bar() {}
+}
+```
+
+While this conflicts with other future work ideas like `const` trait methods or `const trait` declarations,
```
varkor: As far as I can tell, this doesn't conflict: these two syntaxes complement each other. Allowing default `const` method definitions is strong motivation for allowing `const` on methods (and `const trait` in general) from the beginning in my view.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,496 @@
...
+
+## `const` default method bodies
+
+Trait methods can have default bodies for methods that are used if the method is not mentioned
+in an `impl`. This has several uses, most notably
+
+* reducing code repetition between impls that are all the same
+* adding new methods is not a breaking change if they also have a default body
+
+In order to keep both advantages in the presence of `impl const`s, we need a way to declare the
```
varkor: It would be good to mention that non-`const` default methods are no longer a non-breaking API change (but adding `const` default methods are).

const-generic-const-fn-bounds.md
oli-obk: > Also, even without directly accessing the parameters, we have things like:

That is impossible, the `t` itself is a runtime value, you can't have a compile-time value depend on a runtime value. `const` items aren't the same thing as immutable let bindings.

> Is there a reason generics cannot be used inside const contexts?

Basically the reason is that we don't have a concept of generic modules, and items declared inside a function body are essentially inside their own anonymous module. It's entirely unrelated to const contexts and all about items. You can't do

```rust
fn foo<T>(t: T) -> i32 {
    fn bar(t: T) -> i32 {
        42
    }
    bar()
}
```

Array lengths are an anonymous const item and thus subject to the same restrictions that other items are

const-generic-const-fn-bounds.md
oli-obk: How would you declare a method whose default body is `const`, but the method is not required to have a `const` body in implementations? This question is the conflict that I mean and I should clarify this.

const-generic-const-fn-bounds.md
oli-obk: It's the other way around, is it not? Adding a non-const default method will require all `impl const`s to implement that method in a `const` way.

oli-obk: I moved this into the main proposal

oli-obk:
>  I think explicit const trait bounds are necessary, [... ] without this ability, some behaviour is impossible without it

Can you give an example of behaviour you want, but only works if trait methods can be `const`?

const-generic-const-fn-bounds.md
varkor: > That is impossible, the t itself is a runtime value, you can't have a compile-time value depend on a runtime value.

Oh, d'uh, of course.

> Basically the reason is that we don't have a concept of generic modules, and items declared inside a function body are essentially inside their own anonymous module.

Okay, so it's an implementation detail more than anything else. In that case, it's not necessary to have `const` bounds until that's possible.

varkor:
> Can you give an example of behaviour you want, but only works if trait methods can be const?

Sorry, to clarify I mean `<T: const Trait>`, which is only necessary if it's possible to use generic parameters inside `const` contexts/items. Until it's possible to do this, such a facility is unnecessary.

const-generic-const-fn-bounds.md
varkor: This is a good question: it comes from the distinction between traits as signatures versus definitions. Can we use specialisation directly for this? I.e. the trait signature is always `const`-free to make it forwards compatible with adding this sugar? So if you want a `const` default implementation that doesn't require implementers to `const`-implement it, then you provide `default const fn` in an `impl`, but don't specify it `const` in the trait signature itself?

varkor:
To clarify my comment https://github.com/rust-rfcs/const-eval/pull/8#discussion_r247891620, I think if you want a `const` default method implementation that _doesn't_ force implementers to `const`-implement that method, you should have to use specialisation:
```rust
trait Trait {
	fn foo();
}

impl<T> Trait for T {
	default const fn foo() { ... }
}
```
The problem with adding `const` to mean const-default in trait signatures is that it's solely an implementation detail, not a signature detail, so it makes what traits represent blurry (this is already true in some respect for default implementations, but this can be explained as syntactic sugar). You can either use `const` in traits to refer to signature or implementation and I think it's important that we keep `const` signatures in traits free for potential use in the future, as they're a really clean way to deal with abstraction.

This means `const default` is dependent on `default` being stabilised first, but this gives us some leeway to play with the design later. I think it would be very unfortunate to lose the opportunity to define `const` signatures in the future, and I think it's unintuitive with respect to the model of a trait.

const-generic-const-fn-bounds.md
oli-obk: Uh... specialization doesn't really seem appropriate here. It's a completely different scheme. We could also go with an attribute

```rust
trait Foo {
    #[default_method_body_is_const]
    fn foo() {}
}
```

I do see how

```rust
trait Foo {
    const fn foo() {}
}
```

could give the image of the function being const and not just the default body being so. So some other scheme should be chosen regardless of whether `const` methods are a thing we'll ever want.

oli-obk:
I fully agree with the motivation, but not with the solution. As long as we are usually specifying (non-const) default method bodies direclty in the trait, we should keep on doing so and do it for default const method bodies, too.

IIRC that's also not how specialization works, you either specialize all methods, or none. So it would be inappropriate to use on existing traits like `Iterator`, because we'd end up with a breaking change.

varkor:
> I fully agree with the motivation, but not with the solution. As long as we are usually specifying (non-const) default method bodies direclty in the trait, we should keep on doing so and do it for default const method bodies, too.

That's definitely the ideal solution. It's just a bit tricky to nail down the syntax.

> IIRC that's also not how specialization works, you either specialize all methods, or none.

I got the syntax wrong in my example (corrected now), but I think you may only specialise some methods (i.e. the ones included in the `default impl`). This is what makes default trait method definitions sugar for specialisation (e.g. as described in [this section of the specialisation RFC](https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md#default-impls)).

const-generic-const-fn-bounds.md
varkor: > Uh... specialization doesn't really seem appropriate here. It's a completely different scheme.

It's supposed to be a desugaring of default trait implementations. (But perhaps not intuitively from a user's POV).

> We could also go with an attribute

I would be happy going with an attribute for the proposal, but possibly leaving it as something to revisit for stabilisation?

oli-obk:
> I got the syntax wrong in my example (corrected now), but I think you may only specialise some methods (i.e. the ones included in the default impl). This is what makes default trait method definitions sugar for specialisation

Neat, I did not know that.

I went with the attribute and added the syntax question to the unresolved questions.

gnzlbg:
One of the alternatives in https://github.com/rust-lang/rfcs/issues/2626 could benefit of enforcing that a trait method is always a `const fn` on all implementations.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,499 @@
...
+
+    ```rust
+    const F: fn() -> i32 = ...;
+    ```
+
+    is already legal in Rust today, even though the `F` doesn't need to be a `const` function.
+
+2. Opt out bounds are ugly
+
+    I don't think it's either intuitive nor readable to write the following
```
RalfJung: This is quite funny given that it behaves exactly like the `?const` trait bound. Why do you think it is intuitive or readable there, but not here?

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,499 @@
...
+impl const Add for MyInt {
+    fn add(self, other: Self) -> Self {
+        MyInt(self.0 + other.0)
+    }
+}
+```
+
+The const requirement is inferred on all bounds of the impl and its methods,
+so in the following `H` is required to have a const impl of `Hasher`, so that
+methods on `state` are callable.
```
RalfJung: This requirement is in place only when the code is called from const context, right?  The text makes it sound like just writing this impl imposes some requirements, but I don't think that is the case.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,499 @@
...
+```
+
+The same goes for associated types' bounds: all the bounds require `impl const`s for the type used
+for the associated type:
+
+```rust
+trait Foo {
+    type Bar: Add;
+}
+impl Foo for A {
```
RalfJung: Did you mean `impl const Foo`?

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,499 @@
...
+    type Bar = B; // B must have an `impl const Add for B`
+}
+```
+
+If an associated type has no bounds in the trait, there are no restrictions to what types may be used
+for it.
+
+These rules for associated types exist to make this RFC forward compatible with adding const default bodies
+for trait methods. These are further discussed in the "future work" section.
```
RalfJung: The rules would also otherwise be needed, would they not? After all, a `const fn` could use these bounds to call methods on data that has the associated type.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,499 @@
...
+
+## Generic `impl` blocks
+
+Similar to generic parameters on `const` functions, one can have generic parameters on `impl` blocks.
+These follow the same rules as bounds on `const` functions:
+
+* all bounds are required to have `impl const` for substituted types if the impl is used in a const context
+    * except in the presence of `?const` (see below)
```
RalfJung: The fact that this exception is first mentioned when it comes to impl blocks makes it sound like it does not apply for what got previously discussed.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,499 @@
...
+compiler will automatically generate `Drop::drop` calls to the fields:
+
+```rust
+struct Foo;
+impl Drop for Foo { fn drop(&mut self) {} }
+struct Bar(Foo);
+impl const Drop for Foo { fn drop(&mut self) {} } // not allowed
+```
+
+## Runtime uses don't have `const` restrictions?
```
RalfJung: Why is this a question?

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,499 @@
...
+
+`impl const` blocks additionally generate impls that are not const if any generic
+parameters are not const.
```
RalfJung: I find it much easier to think about this as a single impl with conditional bounds, rather than copy-and-pasting two impls.  In a "full version" we'd have (boolean-typed) variables indicating whether something is const, and an operator like `const(c) Trait` that means `const Trait` when `c == true`, and `?const Trait` (as in, no requirement) when `c == false`.  (I find the `?const` notation really confusing when talking like this, but agree that it makes sense on the Rust source side and anyway don't have a better proposal...)

So, `impl<T: Add> const Add for Foo<T>` would really be `impl<c: bool, T: const(c) Add> const(c) Add for Foo<T>`. That explains all of the behavior described below.

const-generic-const-fn-bounds.md
```diff
@@ -0,0 +1,499 @@
...
+of writing this RFC) most definitely not const, because `String` only has an `impl Add for String`
+and not an `impl const Add for String`.
+
+This goes in hand with the current scheme for const functions, which may also be called
+at runtime with runtime arguments, but are checked for soundness as if they were called in
+a const context. E.g. the following function may be called as
+`add(String::from("foo"), String::from("bar"))` at runtime.
+
+```rust
+const fn add<T: Add>(a: T, b: T) -> T {
```
RalfJung: Same here, this is sugar for `const(c) fn add<c: bool, T: const(c) Add>`.  (Sorry for the bad scoping, `c` is bound too late but that's how our syntax works.^^  The `const` should really be on the arrow, but whatever.)

const-generic-const-fn-bounds.md
RalfJung: > To me it feels like ?const fn() -> i32 is ?(const fn() -> i32) and not (?const) fn() -> i32.

But you don't interpret `?const Trait` as `?(const Trait)` -- why?

const-generic-const-fn-bounds.md
RalfJung: I find the overall structure here a bit confusing.  You first have some text outside any subsection that seems to give an overview, cutting across a bunch of features, and then you start to systematically explore `const Trait` in some areas -- but there is no subsection specifically for "`const Trait` in function bounds".  Hence this is the first time you are summarizing, but you are summarizing for a *new* area that we didn't look at, and you are also adding a new concept while summarizing.  Now it is unclear how much of this summary also applies to the previously discussed area of bounds for functions.

const-generic-const-fn-bounds.md
oli-obk: Not sure, probably because `const fn` is a pattern burned into my mind by now, any variance on it is unusual.

Actually I've started to read `?const fn()` correctly over time, so it's fine, and mostly orthogonal to this RFC I'd think

const-generic-const-fn-bounds.md
oli-obk: It's not supposed to be a summary, but just a terse explanation of the new feature. Due to the similarity to bounds on `const` functions I wanted to highlight the differences and similarities in a bullet point way.

RalfJung: I think the structure would be improved by extending this subsection to cover generic bounds in both `fn` and `impl` contexts. The rules are the same, after all.

oli-obk:
Thanks everyone for the critical reviews and constructive commentary. I believe this RFC is ready now and I'll open a regular rust RFC some time today

PR #10, opened 2018-10-08
More on references and promotion 

RalfJung:
Note that I am not entirely sure whether these are the right desugarings. They reflect my current mental model, but that model may well be wrong.

### DIFF
--- a/promotion.md
+++ b/promotion.md
@@ -1,5 +1,11 @@
 # Const promotion
 
+"Promotion" is a mechanism that affects code like `&3`: Instead of putting it on
+the stack, the `3` is allocated in global static memory and a reference with
+lifetime `'static` is provided.  This is essentially an automatic transformation
+turning `&EXPR` into `{ const _PROMOTED = &EXPR; EXPR }`, but only if `EXPR`
+qualifies.
+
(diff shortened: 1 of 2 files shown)


promotion.md
```diff
@@ -1,5 +1,11 @@
 # Const promotion
 
+"Promotion" is a mechanism that affects code like `&3`: Instead of putting it on
+the stack, the `3` is allocated in global static memory and a reference with
+lifetime `'static` is provided.  This is essentially an automatic transformation
+turning `&EXPR` into `{ const _PROMOTED = &EXPR; EXPR }`, but only if `EXPR`
```
oli-obk: This is what we used to have, it's `{ static _PROMOTED = EXPR; &EXPR }` now

RalfJung: I was torn between the two. However:

* Notice that together with my changes in `const_refs.md`, that's actually the same thing.
* If what you said was true, `EXPR` would have to be `Sync`.

RalfJung: Also, it seems that promotion inherits the "must not access statics" rule from `const`, so the `static` desugaring is not really accurate.

PR #18, opened 2018-11-25
Show the unconstness in the example functions

oli-obk:
cc @zoxc 

r? @RalfJung 

Even if we haven't decided on a scheme for how to mark things as unconst, we should not give the appearance of allowing unconst things in safe const code

### DIFF
--- a/const_safety.md
+++ b/const_safety.md
@@ -114,13 +114,13 @@ some extensions of the miri engine that are already implemented in miri) even
 though it uses raw pointer operations:
 ```rust
 const fn test_eq<T>(x: &T, y: &T) -> bool {
-    unsafe { x as *const _ == y as *const _ }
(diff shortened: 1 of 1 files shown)


Zoxc:
I think it would be much clearer if you used a hypothetical `unconst` keyword instead of conflating things with memory safety.

oli-obk:
Hmm... @RalfJung on discord @Zoxc made a good point mentioning that we can never know, whether two promoted constants point to the same memory. e.g. `eq(&42, &42)` might yield `true` or `false` depending on the position of Jupiter.

RalfJung:
We better have a "Jupiter" effect in our type system then. :P

But yes, pointer comparison is a really tricky subject. Are you saying we should use another example instead?

oli-obk:
I updated the examples to show an ok use, and a wrong use

Issue #17, opened 2018-11-24
Dynamic checks for all promotion/const soundness concerns?


RalfJung:
With https://github.com/rust-lang/rust/pull/56123, we can bypass const qualification, so we can e.g. write constants that read from interior mutable statics, and probably other things I forgot to think about.

Could we
* implement all these restrictions as dynamic checks (during the miri execution, similar to https://github.com/rust-lang/rust/pull/56007)
* and collect in this repo what these constraints are?

Ideally, these restrictions would be sufficient, so the soundness criterion for the static checks could be fully described as "statically check for the absence of the following miri errors" -- and such that @oli-obk's feature would actually be sound!

For example, I could imagine tweaking the CTFE instance of miri such that reading from a static is disallowed (and not just disallowing writing to a static like now), implementing a dynamic check that would reject code like what I described above.

For the UnsafeCell and Drop checks, we'd need some value-based analysis, which @oli-obk wanted in the long run anyway.

Cc @eddyb

RalfJung:
Recently @eddyb said something like "turns out Miri catches errors in const qualif, that's good news" -- basically referring to this issue. I take this as confirmation that this is a goal worth pursuing.

With https://github.com/rust-lang/rust/pull/58351, a bunch of "holes" in these checks are fixed. But I think I need help to figure out what is left. Here is an incomplete list, with examples that we could add as `-Zunleash-miri` tests:

* We have to check that const/promoteds do not access statics.

    How would we implement this? We could tell CTFE whether it is currently evaluating a `static` or other initializer, and then we could reject reads/writes to/from "global" (already interned) allocations that are marked mutable. That would actually correspond to the more refined analysis that does allow the above code, but does not allow the variant with added interior mutability. Intuitively, if only immutable memory has been written, then clearly the end result must be the same no matter when evaluation happens.

    ```rust
    use std::cell::Cell;

    static STATIC: usize = 42;
    static mut STATIC_MUT: usize = 42;
    static mut STATIC_CELL: Cell<usize> = Cell::new(42);

    const SCARY_BUT_OKAY: usize = STATIC+1;
    const BAD_MUT: usize = unsafe { STATIC_MUT+1 };
    const BAD_CELL: usize = unsafe { *(&STATIC_CELL as *const _ as *const usize as *mut usize) +1 };
    ```

    @eddyb is there any deep reason to forbid `SCARY_BUT_OKAY`?

[...]

PR #27, opened 2019-08-25
Towards dynamic const-qualify

RalfJung:
I finally came around to look at https://github.com/rust-rfcs/const-eval/issues/17 again, and updated the documents with what I learned on the way.

@oli-obk @eddyb please check if this all makes sense. :)

### DIFF
--- a/const.md
+++ b/const.md
@@ -1,12 +1,14 @@
-# Further restrictions for constants
+# Constants
 
-The [const safety](const_safety.md) concerns apply to all computations happening
-at compile-time, `static` and `const` alike.  However, there are some additional
-considerations about `const` specifically.  These arise from the idea that
+"Constants" in this document refers to `const` bodies, array sizes, and non-`Copy` array initializers.
+On top of what applies to [statics](static.md), they are subject to an additional constraint: In code like
 ```rust
 const CONST: T = EXPR;
 ```
 is supposed to behave as-if `EXPR` was written at every use site of `CONST`.
+
+Based on this requirement, we allow other constants and [promoteds](promotion.md) to read from constants.
+This is why the value of a `const` is subject to validity checks.
 
 ## References
 
@@ -23,9 +25,9 @@ const REF: &u32 = { const _VAL = EXPR; static _STATIC = EXPR; &_STATIC };
 (`EXPR` is assigned to a `const` first to make it subject to the restrictions
 discussed in this document.)
 
-There are three reasons why this could be an issue.
+There are various reasons why this could be an issue.
 
-### Pointer equality
+### 1. Pointer equality
 
 We effectively "deduplicate" all the allocations that would otherwise locally be
 created at each use site of `REF`.  This is observable when the program compares
@@ -33,13 +35,18 @@ these pointers for equality.  We consider this okay, i.e., programs may not rely
 on such constants all getting distinct addresses.  They may not rely on them all
 getting the same address either.
 
-### Interior mutability
+### 2. Interior mutability
 
 If the reference has type `&Cell<i32>` it is quite clear that the program can
 easily observe whether two references point to the same memory even without
 comparing their address: Changes through one reference will affect reads through
 the other.  So, we cannot allow constant references to types that have interior
-mutability (types that are not `Freeze`).
+mutability (types that are not `Freeze`):
+
+```rust
+const BAD: &Cell<i32> = &Cell::new(42);
+// Inlining `BAD` everywhere clearly is not the same as them all pointing to the same thing.
+```
 
 However, we can do better than that: Even if a *type* is not `Freeze`, it can
 have *values* that do not exhibit any interior mutability.  For example, `&None`
@@ -47,7 +54,7 @@ at type `&Option<Cell<i32>>` would be rejected by the naive analysis above, but
 is actually accepted by the compiler because we know that there is no
 `UnsafeCell` here that would permit interior mutability.
 
-### `Sync`
+### 3. `Sync`
 
 Finally, the same constant reference is actually shared across threads.  This is
 very similar to multiple threads having a shared reference to the same `static`,
@@ -59,12 +66,9 @@ ecosystem that would break if we just started enforcing this now. See
 [this issue](https://github.com/rust-lang/rust/issues/49206) and the
 [PR attempting to fix this](https://github.com/rust-lang/rust/pull/54424/).
 
-### `Drop`
+### 4. Drop
 
-Values of "needs drop" types
-can only be used as the final initialization value of a `const` or `static` item.
-They may not be used as intermediate values that would be dropped before the item
-were initialized. As an example:
+`Drop` is actually not an issue, at least not more so than for statics:
 
 ```rust
 struct Foo;
@@ -76,18 +80,10 @@ impl Drop for Foo {
 }
 
 const FOO: Foo = Foo; // Ok, drop is run at each use site in runtime code
-static FOOO: Foo = Foo; // Ok, drop is never run
 
 // Not ok, cannot run `Foo::drop` because it's not a const fn
 const BAR: i32 = (Foo, 42).1;
 ```
-
-This restriction might be lifted in the future after trait impls
-may be declared `const` (https://github.com/rust-rfcs/const-eval/pull/8).
-
-Note that in promoteds this restriction can never be lifted, because
-otherwise we would silently stop calling the `Drop` impl at runtime and
-pull it to much earlier (compile-time).
 
 ## Reading statics
 
--- a/const.md
+++ b/const.md
@@ -55,10 +55,11 @@ at type `&Option<Cell<i32>>` would be rejected by the naive analysis above, but
 is actually accepted by the compiler because we know that there is no
 `UnsafeCell` here that would permit interior mutability.
 
-*Dynamic check.* The Miri engine could check this dynamically by ensuring that
-the new data that is interned for a constant is all marked as
-immutable. (Constants referring to already existing mutable data are not
-inherently problematic.)
+*Dynamic check.* The Miri engine enforces this dynamically by ensuring that the
+new data that is interned for a constant is all marked as immutable. However,
+note the FIXME added [by this PR](https://github.com/rust-lang/rust/pull/63955):
+for untyped data in a constant, we currently just *make* it immutable, instead
+of checking properly.
 
 ### 3. `Sync`
 
--- a/const.md
+++ b/const.md
@@ -54,6 +54,11 @@ at type `&Option<Cell<i32>>` would be rejected by the naive analysis above, but
 is actually accepted by the compiler because we know that there is no
 `UnsafeCell` here that would permit interior mutability.
 
+*Dynamic check.* The Miri engine could check this dynamically by ensuring that
+the new data that is interned for a constant is all marked as
+immutable. (Constants referring to already existing mutable data are not
+inherently problematic.)
+
 ### 3. `Sync`
 
 Finally, the same constant reference is actually shared across threads.  This is
@@ -65,6 +70,8 @@ However, this does not currently happen, and there are several crates across the
 ecosystem that would break if we just started enforcing this now. See
 [this issue](https://github.com/rust-lang/rust/issues/49206) and the
 [PR attempting to fix this](https://github.com/rust-lang/rust/pull/54424/).
+
+*Dynamic check.* It is unclear how the Miri engine could dynamically check this.
 
 ### 4. Drop
 
@@ -97,3 +104,6 @@ This is distinct to the concern about interior mutability above: That concern
 was about first computing a `&Cell<i32>` and then using it at run-time (and
 observing the fact that it has been "deduplicated"), this here is about using
 such a value at compile-time even though it might be changed at run-time.
+
+*Dynamic check.* The Miri engine could check this dynamically by refusing to
+access mutable global memory when computing a const.
--- a/const.md
+++ b/const.md
@@ -6,6 +6,7 @@ On top of what applies to [statics](static.md), they are subject to an additiona
 const CONST: T = EXPR;
 ```
 is supposed to behave as-if `EXPR` was written at every use site of `CONST`.
+To make this work, we need to ensure [const safety](const_safety.md).
 
 Based on this requirement, we allow other constants and [promoteds](promotion.md) to read from constants.
 This is why the value of a `const` is subject to validity checks.
--- a/promotion.md
+++ b/promotion.md
@@ -72,7 +72,7 @@ everything, so the only possible remaining failure are panics.
 
 However, things get more tricky when `const` and `const fn` are involved.
 
-For `const`, based on the const safety check described [here](const_safety.md),
+For `const`, based on the const safety check described [here](const_safety.md#const-safety-check-on-values),
 we can rely on there not being const-unsafe values in the `const`, so we should
 be able to promote freely.  For example:
 
@@ -98,6 +98,9 @@ another error (unsupported operation or undefined behavior), we have no choice
 but to abort compilation of a program that would have compiled fine if we would
 not have decided to promote.  It is the responsibility of `foo` to not fail this
 way when working with const-safe arguments.
+
+For this reason, only `const fn` that were explicitly marked with the
+`#[rustc_promotable]` attribute are subject to promotion.
 
 *Dynamic check.* The Miri engine already dynamically detects const safety
 violations, but the main point of promoteds is ruling them out statically.
--- a/promotion.md
+++ b/promotion.md
@@ -142,7 +142,6 @@ promotion in this situation: all function calls are promoted, not just
 Promotion is *not* involved in something like this:
 
 ```rust
-#![feature(const_vec_new)]
 const EMPTY_BYTES: &Vec<u8> = &Vec::new();
 
 const NESTED: &'static Vec<u8> = {
--- a/promotion.md
+++ b/promotion.md
@@ -1,14 +1,20 @@
 # Const promotion
 
-"Promotion" is a mechanism that affects code like `&3`: Instead of putting it on
-the stack, the `3` is allocated in global static memory and a reference with
-lifetime `'static` is provided.  This is essentially an automatic transformation
-turning `&EXPR` into `{ const _PROMOTED = &EXPR; EXPR }`, but only if `EXPR`
-qualifies.
+["(Implicit) Promotion"][rfc] is a mechanism that affects code like `&3`:
+Instead of putting it on the stack, the `3` is allocated in global static memory
+and a reference with lifetime `'static` is provided.  This is essentially an
+automatic transformation turning `&EXPR` into `{ const _PROMOTED = &EXPR; EXPR
+}`, but only if `EXPR` qualifies.
 
 Note that promotion happens on the MIR, not on surface-level syntax.  This is
 relevant when discussing e.g. handling of panics caused by overflowing
 arithmetic.
+
+On top of what applies to [consts](const.md), promoteds suffer from the additional issue that *the user did not ask for them to be evaluated at compile-time*.
+Thus, if CTFE fails but the code would have worked fine at run-time, we broke the user's code for no good reason.
+That's why we have to be very conservative with what can and cannot be promoted.
+
+[rfc]: https://github.com/rust-lang/rfcs/blob/master/text/1414-rvalue_static_promotion.md
 
 ## Rules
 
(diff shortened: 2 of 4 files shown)


RalfJung:
Also Cc @ecstatic-morse

const.md
```diff
@@ -1,13 +1,15 @@
-# Further restrictions for constants
+# Constants
 
-The [const safety](const_safety.md) concerns apply to all computations happening
-at compile-time, `static` and `const` alike.  However, there are some additional
-considerations about `const` specifically.  These arise from the idea that
+"Constants" in this document refers to `const` bodies, array sizes, and non-`Copy` array initializers.
+On top of what applies to [statics](static.md), they are subject to an additional constraint: In code like
```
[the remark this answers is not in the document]
RalfJung: Well, for statics the const safety requirements actually do not matter very much, as I tried to explain in that page.
But I agree that there should be a link to the const safety page.

promotion.md
```diff
@@ -111,6 +118,41 @@ it is unlikely to be the desired behavior in most cases and very likey to be con
...
+    // This does not work when we have an inner scope:
+    let x = &Vec::new(); //~ ERROR: temporary value dropped while borrowed
+    x
+};
+```
+
+In `EMPTY_BYTES`, the reference obtains the lifetime of the "enclosing scope",
+similar to how `let x = &mut x;` creates a reference whose lifetime lasts for
+the enclosing scope. This is decided during MIR building already, and does not
+involve promotion.
```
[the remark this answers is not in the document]
RalfJung: > I don't know how it should extend to arbitrary cases. Is it the braces that cause the problem or the assignment to x?

I don't know. I am paraphrasing @eddyb here.
@eddyb are these rules explained anywhere?

> It also might be good to mention #[rustc_promotable] here, since without it calls to const fns can't get promoted.

Good point, will do.

const.md
```diff
@@ -23,31 +26,41 @@ const REF: &u32 = { const _VAL = EXPR; static _STATIC = EXPR; &_STATIC };
...
+```
 
 However, we can do better than that: Even if a *type* is not `Freeze`, it can
 have *values* that do not exhibit any interior mutability.  For example, `&None`
 at type `&Option<Cell<i32>>` would be rejected by the naive analysis above, but
 is actually accepted by the compiler because we know that there is no
 `UnsafeCell` here that would permit interior mutability.
 
-### `Sync`
+*Dynamic check.* The Miri engine could check this dynamically by ensuring that
```
oli-obk: I thought we were doing this nowadays?

promotion.md
```diff
@@ -111,6 +121,41 @@ it is unlikely to be the desired behavior in most cases and very likey to be con
...
+Promotion is also responsible for making code like this work:
+
+```rust
+const FOO: &'static i32 = {
+    let x = &13;
+    x
+};
+```
+
+However, since this is in explicit const context, we could be less strict about
```
oli-obk: We are less strict. You can promote any const fn call inside a constant

const.md
RalfJung: Not that I know of. I did not have any time recently to work on https://github.com/rust-lang/rust/issues/49206.

promotion.md
RalfJung: Ah, good to know! Are there any other differences? We could in principle be even more aggressive.

oli-obk: iirc we are running just the regular const checks on promoteds inside constants and const fns. So there's no third system, it's exactly as if you wrote an explicit constant

const.md
oli-obk: we have https://github.com/rust-lang/rust/blob/0444b9f66acb5da23dc816e0d8eb59623ba9ea50/src/librustc_mir/interpret/intern.rs#L183

promotion.md
RalfJung: Isnt promotion also about identifying which parts of the MIR to promote, and then actually patch the MIR to replace uses of the data by a promoted? Somehow this still needs to do something extra inside constants I think, or will it just trigger at every `&`?

const.md
RalfJung: Oh sorry, I saw the `Sync` and thought this was about `Sync`. But that's a red line in the diff.

Yes, I think we might be already doing this, but I'd like to review the code specifically checking for this (and maybe add some comments saying why this is important) before making claims here.

promotion.md
oli-obk: So implementation wise the difference is `IsPromotable` vs `IsImplicitlyPromotable`, where iirc `IsImplicitlyPromotable` only makes a difference in non-const fns

RalfJung: So `IsPromotable` is *not* about the kind of promotion I am talking about here? :(
What is it about then?

I asked @eddyb about this terminology but must have misunderstood what he said.

oli-obk: > So IsPromotable is not about the kind of promotion I am talking about here? :(

I'm confused.

So `IsPromotable` and `IsImplicitlyPromotable` are required in normal functions, while `const fn` and constants only check `IsPromotable`

RalfJung: I guess I am confused by the "implicit" part. The promotion to a separate static is also implicit inside a const body.

If the thing inside a normal `fn` is "implicit promotion", I would expect the thing inside a const body to be "explicit promotion", but that does not make sense, does it?

oli-obk: yea... it's not the best naming scheme... Maybe we should call the thing in non-const fns "restricted promotion" and the other cases just "promotion" or "unrestricted promotion"

RalfJung: That doesn't convey why it is restricted, though. "implicit" did that nicely, but was not precise enough.

"promotion of run-time code" is verbose but informative. "run-time promotion" is somewhat misleading... "run-time code promotion"?

RalfJung:
@eddyb
> Promotion in consts and statics is performed by removing the StorageDead - that's what the bitset returned by mir_const_qualif is for - since that already has to work.

So I was using the term for the mechanism, and you were using for the effect. That would explain the confusion.
Nuking `StorageDead` actually makes sense to me, and that's what I used to think happens in consts -- but then I talked with you and you taught me about the "enclosing scope" rule and since then I was sure it was *not* done by nuking `StorageDead`.

I think "promotion" makes more sense as the name for the mechanism of extracting part of the MIR out -- and that's what `Promoted` already means in the compiler. So I feel like we should come up with another name for the effect of something having a longer lifetime than expected. But we can also do it the other way around, as long as we are able to clearly distinguish the two.

Overall we are looking for three terms:
* The mechanism that extracts parts of a MIR and moves it out (the kind of thing `promoted_mir` gives access to, from what I understand).
* The specific case of using that mechanism on unsuspecting run-time code, which has the effect of moving it to compile-time, leading to a very restrictive analysis (this is what `promoted.md` in this repo talks about).
* The analysis that determines which things get `'static` lifetime even though they usually wouldn't. The result of this analysis is used to nuke `StorageDead` in `const` (and `static`?), and to extract part of a MIR and cut it out elsewhere.

Finally I feel like we are making progress on this minefield of three different things that are all referred to by the same name.^^

Strawman proposals for the three names:
* Mechanism: promotion; unsuspecting run-time case: run-time-code promotion; analysis: lifetime extension
* Mechanism: MIR extraction; unsuspecting run-time case: run-time-code promotion; analysis: promotion
* Mechanism: MIR extraction; unsuspecting run-time case: restricted promotion; analysis: promotion

eddyb:
Nuking `StorageDead` isn't enough for non-lifetime things (like `[expr; n]`), and it may be dubious even for the `'static` effect, it's a microoptimization, we should measure how expensive the correct way is.

The reason I was abusing the word "promotion" is mostly historical: the lifting was "merely" the way it was implemented in MIR, but both "qualification" and "promotion" predate MIR, and they were referring to AST expressions: "qualification" was the analysis, and "promotion" was the *effect* of having `'static` lifetime (as observed by borrowck and implemented in AST-based codegen).

The way it was done back then made even something as simple as `1 + 1` behave like `*&1 + *&1`, with some custom constant-folding for reading from a custom allocation (because otherwise the LLVM IR would've been loading *every* little literal from some global).

I should've changed the terminology when I implemented the MIR side, but I guess I was trying to make them as close as possible to avoid introducing any bugs (since the old borrowck would still use the information from the AST pass, until we switched to the MIR borrowck, and only recently are we getting close to getting rid of the old one).

For the MIR, I think "const checking" and "const promotion" are pretty okay terms. One thing your nomenclature seems to avoid is that the analysis used to nuke `StorageDead` (and which should actually be used to do the same extraction as in `fn`s) is the same as the runtime promotion, just more relaxed.

The reason you can be more relaxed is the code *must* run at compile-time anyway, so there is no risk of a `const fn` changing behavior from runtime to compile-time (and breaking compilation). That's pretty much it, the rules around drops and frozen memory are the same.

oli-obk:
The mechanism is called promotion because it is what does the action of promoting.

The unsuspecting-runtime-code-gets-const-evaluated thing should probably focus on the fact that const evaluating arbitrary runtime code will fail, so we need to be careful about what we guarantee. Maybe "guaranteed" is the right keyword?

I like "lifetime extension", because it makes it so much clearer that the type system process could just as well extend the lifetime on `&read_from_network()` and that would be :boom: 	out of obvious reasons.

RalfJung:
> Nuking StorageDead isn't enough for non-lifetime things (like [expr; n]), and it may be dubious even for the 'static effect, it's a microoptimization, we should measure how expensive the correct way is.

Fair. The focus in this PR anyway is to document current behavior and clarify terminology.

> One thing your nomenclature seems to avoid is that the analysis used to nuke StorageDead (and which should actually be used to do the same extraction as in fns) is the same as the runtime promotion, just more relaxed.

Right, so those would be two different "modes" of the analysis, whatever we end up calling that one.

> The mechanism is called promotion because it is what does the action of promoting.

That doesn't exactly explain which of the multiple possible meanings of this verb you mean. ;)

Like, do you consider nuking `StorageDead` as "how promotion looks like in consts", or as *not* being promotion?

I think my preferred proposal is to use the term "promotion" in the same sense as [`StaticKind::Promoted`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/mir/enum.StaticKind.html#variant.Promoted). So that would mean that "promotion" is "messing with MIR to extract some part of it into a separate static". Then we'd have to find a separate term for the analysis, at least right now where it is the case that the analysis does *not* use promotion for all the things it finds (but uses "nuking StorageDead" for some).

Not sure if a separate name for the analysis still makes sense though *if* we end up always using the "messing with MIR" mechanism for everything? Still, seems nice to be able to talk about them separately. It's just that someone would have to rename everything promotion-analysis-related in rustc...

ecstatic-morse:
FWIW, I've been using the following nomenclature in my head while working on rust-lang/rust#63812 and friends. 


- **lifetime extension**: The ultimate goal of RFC #1414. 
- **promotion**: the act of mutating the `mir::Body` to extend the lifetime of certain temporaries. This is narrower than the way this term was being used in the discussion around RFC #1414, since it is merely the mechanism by which lifetime extension is achieved.
- **`fn` promotion**: The specific method of promotion used for `fn` and `const fn`s. This is the typical method of promotion.
 - **`StorageDead` suppression**: A more efficient method of promotion, used in `const`s and `static`s but not `fn`s, that requires fewer modifications to the MIR.
- **promotable**: Whether a temporary is eligible for promotion, and thus to have its lifetime extended.
- **implicitly promotable**: Whether a temporary is eligible for promotion if the user did **not** explicitly ask for its lifetime to be extended by passing it to a function with `#[rustc_args_promotable]`.
- **explicitly promotable**: Whether a temporary is eligible for promotion if the user explicitly asked for its lifetime to be extended by passing it to a function with `#[rustc_args_promotable]`.

With that in mind, the three jobs of the current qualifcation pass are:
- validation (or checking)
- promotable temp identification
- promotion

The last step, **promotion**, has two different subroutines depending on whether the current item is a `const`/`static` or `fn`/`const fn`.

Also, I think **lifetime extension** might already have a specific meaning in the context of the borrow checker?

RalfJung:
@ecstatic-morse thanks a lot, that is very helpful!

> by passing it to a function with #[rustc_args_promotable]

Does that attribute have the same effect as being inside the body of a `const`? Promotability (not just promotion) also works differently there than in the body of a normal `fn`.

> promotable: Whether a temporary is eligible for promotion

What exactly is a "temporary" here? Is it just any MIR-level `Local`?

> I suppose I should be referring to the last two steps together as "lifetime extension", although I mostly just say "promotion" when writing about them. Also, I think "lifetime extension" might already have a specific meaning in the context of the borrow checker?

Call them "promotion" unfortunately overlaps with the third step which is also called "promotion". But what I meant by "lifetime extension" was just what you call "promotability (analysis)".

Maybe using the same name here for both is okay, as long as everyone is aware of the more refined terminology and we switch to it when it seems like confusion is arising.

ecstatic-morse:
> Does that attribute have the same effect as being inside the body of a const? Promotability (not just promotion) also works differently there than in the body of a normal fn.

Ah, that's a good catch. I think [this](https://github.com/rust-lang/rust/blob/f504e37dc64b2f00291f53517fe9cb2329d00fb8/src/librustc_mir/transform/qualify_consts.rs#L1424) is the place where they diverge? Basically, promotion of `#![rustc_args_required_const]` arguments is mandatory in `fn`s and `const fn`s--an error is thrown if the argument is not explicitly promotable, but not mandatory in `const`s and `static`s--since everything inside a `const` body is also `const`.

> What exactly is a "temporary" here? Is it just any MIR-level Local?

A temporary is a MIR `Local` that does not correspond to a named variable in the source. We don't promote named variables because it would surprise the user when their drop impl is not called.

> Call them "promotion" unfortunately overlaps with the third step which is also called "promotion". But what I meant by "lifetime extension" was just what you call "promotability (analysis)".

Ah, I like "promotability analysis" (I didn't have a good noun for the second step). The hierarchy currently looks like this in my head:

- const validation
- lifetime extension (if we can agree on this term)
    - promotability analysis
    - promotion

static.md
```diff
@@ -0,0 +1,36 @@
...
+Statics (`static`, `static mut`) are the simplest kind of compile-time evaluated data:
+* The user explicitly requested them to be evaluated at compile-time,
+  so evaluation errors from computing the initial value of a static are no concern
+  (in other words, [const safety](const_safety.md) is mostly not an issue).
+* They observably get evaluated *once*, with the result being put at some address known at run-time,
+  so there are no fundamental restrictions on what statics can do.
+* The compiler checks that statics are `Sync`, justifying sharing their address across threads.
+* [Constants](const.md) and [promoteds](promotion.md) are not allowed to read from statics,
+  so their final value does not have have to be [const-valid](const_safety.md#const-safety-check-on-values) in any meaningful way.
+  As of 2019-08, we do check them for validity anyway, to be conservative; and indeed constants could be allowed to read from frozen statics.
```
Centril: I think this would be nice personally; less times to invent an ephemeral constant just to please the check and you also get to use exported statics from other crates.

Any future compat hazards with some new feature?

Centril:
> Basically, promotion of `#[rustc_args_required_const]` arguments is mandatory in `fn`s and `const fn`s--an error is thrown if the argument is not explicitly promotable, but unnecessary in `const`s and `static`s--everything inside a `const` body is also `const` and there's no need to extend the lifetime of the argument by removing `StorageDead`.

Question: what is the difference between `#[rustc_args_required_const]` and `const A: B` generic parameters in terms of what code can be passed as an argument?

oli-obk:
no difference for the caller except in syntax.

The function body can't use the argument as const, so you can't forward it internally, so const generics are more powerful there.

static.md
Centril: Discovered a problem... If we want to have a scheme like https://github.com/rust-lang/rfcs/pull/2492 but centered around `static`s instead, allowing `static`s in const contexts will give us all the problems re. separate compilation.

RalfJung:
> Said another way, when it comes to #[rustc_args_required_const] arguments (and probably array initializers), the purpose of promotion is not actually lifetime extension since there's no reference involved.

I see. That sounds quite hacky. But on the other hand we probably do want the same checks for both cases (as in both cases it is not locally syntactically apparent that things are evaluated at const-time). So maybe this is a hint that the *analysis* (shared by promotion-in-runtime-fn and `rustc_args_required_const`) should be called something that does not involve the term "promotion", as it is not exclusive to promotion.

Maybe it makes more sense then to to speak of "explicit-const validation" and "implicit-const validation" or "explicit-const checking" and "implicit-const checking" (I think I prefer the latter) to refer to the analysis that checks if code is amenable to const-eval. And then both `rustc_args_required_const` and promotion-in-runtime-fn would be using the results "implicit-const checking". (The brackets here are like "(implicit-const) checking" as the check is about testing whether this code can be implicitly treated as being const-code.)

RalfJung:
It seems like we are mostly bikeshedding terminology here at this point, which IMO does not have to block this PR. So, maybe we can move that to an issue? Is there anything else blocking the PR?

ecstatic-morse:
I'm still confused about [the first part of this comment](https://github.com/rust-rfcs/const-eval/pull/27#discussion_r317931570). Other than that I think it's ready.

promotion.md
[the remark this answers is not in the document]
RalfJung: Hm I guess I should edit this in the light of our unresolved terminology.

Is the entire "Promotion is *not* involved" incorrect, or is there something there that makes sense? I was trying to explain why `const EMPTY_BYTES: &Vec<u8> = &Vec::new();` works even though this stuff does *not* get promoted. @ecstatic-morse since my explanation confused you, could you try to explain this in your own words?

static.md
RalfJung: @Centril sorry, I do not follow... why is it harder for separate compilation to support `const` referring to `static` than `const` referring to `const` (and the latter we already support)?

promotion.md
```diff
@@ -111,6 +121,41 @@ it is unlikely to be the desired behavior in most cases and very likey to be con
...
+```rust
+const FOO: &'static i32 = {
+    let x = &13;
+    x
+};
+```
+
+However, since this is in explicit const context, we are less strict about
+promotion in this situation: all function calls are promoted, not just
+`#[rustc_promotable]` functions.
```
ecstatic-morse: @RalfJung  I think my confusion really occurs because of this paragraph. It seems to apply that *both* `const`s should compile in the following example, which is not true.

promotion.md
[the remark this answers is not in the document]
ecstatic-morse: See above comment for what is actually confusing me here. The "enclosing scope" rule makes sense intuitively, but the details are still not clear to me. That's my problem though, not this PR's.

oli-obk:
r=me with @ecstatic-morse's confusion resolved

promotion.md
RalfJung: Ah yes. That's because of the `Drop` check, which still applies.

promotion.md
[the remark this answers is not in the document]
RalfJung: I have done some edits. @ecstatic-morse does this resolve your confusion?

promotion.md
ecstatic-morse: It also fails on types [which are not `Drop`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=30be96601b71d20a232d99e610e394d4). I think @oli-obk comment above was incorrect?

RalfJung: That example [works](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=92003e9d833ac8dc9f6b63eb6720aa8b) if you make `new` a `const fn`.

ecstatic-morse:
My confusion has been resolved (at least far as this PR is concerned).

RalfJung:
Great, thanks. :)

promotion.md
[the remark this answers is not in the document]
eddyb: The enclosing scope rule is the same one that makes these two different:
```rust
{
    let r = &foo(); // compiles
    bar(r);
}
```
```rust
{
    let r = id(&foo()); // doesn't compile
    bar(r);
}
```
Except instead of `let` it's `const`/`static` and instead of "enclosing block" it's `'static`.

I don't remember the exact name of these semantics, maybe "rvalue/temporary lifetime/scope"? @nikomatsakis would be more helpful here, sorry.

eddyb:
Note to self: review this (eventually).

promotion.md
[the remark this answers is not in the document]
nikomatsakis: I call these "temporary lifetimes" -- I've been meaning to writeup some text about them for the rust reference.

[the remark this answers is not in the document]
nikomatsakis: I call these "temporary lifetimes" -- I've been meaning to writeup some text about them for the rust reference.

static.md
Centril: In this case it would be a `static` for which the value isn't known yet; that is something like `pub extern static Foo: u8;` <-- no value. That would be similar to https://github.com/rust-lang/rfcs/pull/2492 in that it is matched up with something in the crate graph. If you can then refer to `Foo` in types then that will create all the same problems as with `extern existential type`.

RalfJung: As long as whatever *instantiates* the `static` checks them for validity, we are sill good. The important part is the invariant that all statics have been checked. The order doesn't matter.

PR #28, opened 2019-10-11
Define promotion contexts and promotability

ecstatic-morse:
This adds an introduction to `promotion.md` that explains promotion in terms of "promotion contexts" (e.g. `&expr`, `[expr; 32]`) and "promotability" (the rules for which expressions can be promoted in a given context). This is a variation on the terminology described [in this comment](https://github.com/rust-rfcs/const-eval/pull/27#issuecomment-531953305), which I was using to reason about the code in `qualify_consts.rs` while working on dataflow-based const qualification.

I'm not asserting that this mental model is the correct one. In particular, you can see some inconsistencies in the section called "Lifetime extension in `const` and `static`". Hopefully it can serve as a starting point.

This also adds some requirements that were missing from the "Rules" section (now renamed to "Promotability"). These requirements are mostly structural ones (single assignment, no named variables), so maybe they should go in a different section?

cc @RalfJung @oli-obk

### DIFF
--- a/promotion.md
+++ b/promotion.md
@@ -28,10 +28,11 @@ first context where promotion was done.
 
 ### Non-`Copy` array initialization
 
-Another promotion context was introduced in [RFC
-2203](https://github.com/rust-rfcs/const-eval/blob/master/const.md). In this
-case, we try to promote the initializer in expressions like `[Vec::new(); 32]`,
-which allows non-`Copy` types to be used as array initializers.
+Another promotion context was introduced in [RFC 2203][]. In this case, we try
+to promote the initializer in expressions like `[Vec::new(); 32]`, which allows
+non-`Copy` types to be used as array initializers.
+
+[RFC 2203]: https://github.com/rust-lang/rfcs/blob/master/text/2203-const-repeat-expr.md
 
 ### `#[rustc_args_required_const(...)]`
 
--- a/promotion.md
+++ b/promotion.md
@@ -72,16 +72,38 @@ implicit context are a superset of the ones in an explicit context.
 
 [warn-rfc]: https://github.com/rust-lang/rfcs/blob/master/text/1229-compile-time-asserts.md
 
-### Promotion contexts in `const` and `static`
+### Promotion contexts inside `const` and `static`
+
+Lifetime extension is also responsible for making code like this work:
+
+```rust
+const FOO: &'static i32 = {
+    let x = &13;
+    x
+};
+```
 
 We defined above that promotion guarantees that code in a non-const context
-will be executed at compile-time. However, lifetime extension and non-`Copy`
-array initialization are useful features *inside* `const`s and `static`s as
-well. Strictly speaking, the transformation used to enable these features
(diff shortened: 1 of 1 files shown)


RalfJung:
This needs rebasing now that my PR landed.

ecstatic-morse:
Rebased. There's now two sections about `&` in `const`s and `static`s. I will resolve this once the terminology is decided, since the section I added might go away.

promotion.md
```diff
@@ -10,17 +20,113 @@ Note that promotion happens on the MIR, not on surface-level syntax.  This is
...
+Another promotion context, the initializer of an array expression, was
+introduced in [RFC 2203][]. Here, promotion allows arrays of
+non-`Copy` types to be initialized idiomatically, for example
+`[Option::<Box<i32>>::None; 32]`.
+
+[RFC 2203]: https://github.com/rust-lang/rfcs/blob/master/text/2203-const-repeat-expr.md
+
+### `#[rustc_args_required_const(...)]`
+
+Additionally, some platform intrinsics require certain operations to be
```
ecstatic-morse:
```suggestion
Additionally, some platform intrinsics require certain parameters to be
```

promotion.md
```diff
@@ -10,17 +20,113 @@ Note that promotion happens on the MIR, not on surface-level syntax.  This is
...
+implicit context are a superset of the ones in an explicit context.
+
 [warn-rfc]: https://github.com/rust-lang/rfcs/blob/master/text/1229-compile-time-asserts.md
 
-## Rules
+### Promotion contexts in `const` and `static`
+
+We defined above that promotion guarantees that code in a non-const context
+will be executed at compile-time. However, lifetime extension and non-`Copy`
+array initialziation are useful features *inside* `const`s and `static`s as
```
ecstatic-morse:
```suggestion
array initialization are useful features *inside* `const`s and `static`s as
```

promotion.md
```diff
@@ -10,17 +20,113 @@ Note that promotion happens on the MIR, not on surface-level syntax.  This is
...
+```
+
+Once again, this is not a fundamental limitation in the CTFE engine; we are
+perfectly capable of evaluating such expressions at compile time. However,
+determining the promotability of complex expressions would require more
+resources for little benefit.
+
+### Access to a `const` or `static`
+
+Accesses to `const`s are always promotable, regardless of the body of the
```
oli-obk: well, not quite regardless. You still can't promote `Cell`s

oli-obk:
lgtm other than the clarification about which constants get promoted

promotion.md
```diff
@@ -1,6 +1,16 @@
 # Const promotion
 
-["(Implicit) Promotion"][promotion-rfc] is a mechanism that affects code like `&3`:
+"Promotion" is the act of guaranteeing that code not written in a const context
+(e.g. initalizer of a `const` or `static`, or an array length expression) will
```
RalfJung: These are examples *for* code in const-context, right? Because to me this reads like examples for "code not written in a const context".

Issue #20, opened 2018-12-14
Heap allocations in constants


oli-obk:
# Motivation

In order to totally outdo any other constant evaluators out there, it is desirable to allow things like using serde to deserialize e.g. json or toml files into constants. In order to not duplicate code between const eval and runtime, this will require types like `Vec` and `String`. Otherwise every type with a `String` field would either need to be generic and support `&str` and `String` in that field, or just outright have a mirror struct for const eval. Both ways seem too restrictive and not in the spirit of "const eval that just works".

# Design

## Allocating and Deallocating

Allow allocating and deallocating heap inside const eval. This means `Vec`, `String`, `Box`
    * Similar to how `panic` is handled, we intercept calls to an [allocator's `alloc`](https://doc.rust-lang.org/std/alloc/trait.Alloc.html#tymethod.alloc) method and never actually call that method. Instead the miri-engine runs const eval specific code for producing an allocation that "counts as heap" during const eval, but if it ends up in the final constant, it becomes an unnamed static. If it is leaked without any leftover references to it, the value simply disappears after const eval is finished. If the value is deallocated, the call to `dealloc` in intercepted and the miri engine removes the allocation. Pointers to dead allocations will cause a const eval error if they end up in the final constant.

## Final values of constants and statics

If a constant's final value were of type `String`, and the string is not empty, it would be very problematic to use such a constant:

```rust
const FOO: String = String::from("foo");
let x = FOO;
drop(x);
// how do we ensure that we don't run `deallocate`
// on the pointer to the unnamed static containing the bye sequence "foo"?
```

While there are a few options that could be considered, all of them are very hard to reason about and easy to get wrong. I'm listing them for completeness:

* just set the capacity to zero during const eval
    * will prevent deallocation from doing anything
    * seems like it would require crazy hacks in const eval which know about types with heap allocations inside
    * Not sure how that would work for `Box`
* use a custom allocator that just doesn't deallocate
    * requires making every single datastructure generic over the allocator in use
    * doesn't fit the "const eval that just works" mantra
* actually turn const eval heap allocations into real heap allocations on instantiation
    * not zero cost
    * use of a constant will trigger a heap allocation

We cannot ban types that contain heap allocations, because

```rust
struct Foo;

impl Drop for Foo {
    fn drop(&mut self) {
        println!("foo");
    }
}

const FOO: Foo = Foo;
```

[...]

PR #29, opened 2019-11-03
state explicitly that const-items are copied bitwise

RalfJung:
This certainly describes the current reality.

Cc @ecstatic-morse

### DIFF
--- a/const.md
+++ b/const.md
@@ -6,6 +6,7 @@ On top of what applies to [statics](static.md), they are subject to an additiona
 const CONST: T = EXPR;
 ```
 is supposed to behave as-if `EXPR` was written at every use site of `CONST`.
+However, in reality `EXPR` is just computed once and the result is bit-wise copied each time `CONST` is used.
 To make this work, we need to ensure [const safety](const_safety.md).
 
 Based on this requirement, we allow other constants and [promoteds](promotion.md) to read from constants.
--- a/const.md
+++ b/const.md
@@ -6,7 +6,7 @@ On top of what applies to [statics](static.md), they are subject to an additiona
 const CONST: T = EXPR;
 ```
 is supposed to behave as-if `EXPR` was written at every use site of `CONST`.
-However, in reality `EXPR` is just computed once and the result is bit-wise copied each time `CONST` is used.
+However, in current versions of rustc, `EXPR` is just computed once and the result is bit-wise copied each time `CONST` is used.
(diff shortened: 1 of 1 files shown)


ecstatic-morse:
As I noted on Zulip, I think we should leave this deliberately unspecified until we have a strategy for [heap allocation in constants](https://github.com/rust-rfcs/const-eval/issues/20#issuecomment-520227200). I don't think we gain much by specifying this today? Although I guess this repo is non-normative, so explaining this as the *current* behavior would be fine.

Basically, the issue is around code like:

```rust
const ONE: Box<i32> = Box::new(1);

let a = ONE;
let b = ONE;
```

My intuitive understanding of `const` is that it behaves as though you copy-pasted the initializer everywhere the `const` was used, in which case this should compile. However, if bitwise-copies are used everywhere, obviously this code obviously must be rejected.

RalfJung:
> I don't think we gain much by specifying this today

I think we gain a lot in clarity.^^ The only reason this sentence isn't there already is that when writing this document, I considered it so obvious that it wasn't worth stating...

> My intuitive understanding of const is that it behaves as though you copy-pasted the initializer everywhere the const was used

It behavs *as-if* but without actually doing that, that's the entire point. `const` would be rather useless if it was actual copy-paste... or at least the name would be quite the misnomer. There is nothing *constant* about them any more!

ecstatic-morse:
My understanding is that specifying that a `const` always uses bitwise copies means that, under *all* possible proposals for heap allocation in constants, the above example *must* fail to compile. Is this correct?

If so, then I think this should at least be discussed in rust-rfcs/const-eval#20 before a decision is made here. I'm not convinced that the previous example must be accepted, but I don't think it's outside the realm of possibility.

RalfJung:
I expected that to be a constraint anyway, but I'll add a clause clarifying that this is current behavior.
And in any case, nothing in this repo is normative -- these are just docs.

RalfJung:
@ecstatic-morse e.g., everything [we say here](https://github.com/rust-rfcs/const-eval/blob/master/const.md#references) is also just describing current behavior and could be changed if we wanted to.

ecstatic-morse:
True. In any case, we should try to reach consensus about this when @oli-obk is back from vacation.

I would r+ the new changes if I had the permissions btw.

RalfJung:
> I would r+ the new changes if I had the permissions btw.

That sounds good enough for such a small change. ;)

PR #30, opened 2019-11-06
Replace `rust-rfcs` with `rust-lang`

JohnTitor:


oli-obk:
heh, good catch, thanks!

PR #31, opened 2019-11-16
promotion tweaks

RalfJung:
r? @oli-obk @ecstatic-morse

### DIFF
--- a/promotion.md
+++ b/promotion.md
@@ -232,15 +232,17 @@ restrictions described above still apply for the *result* of the promoted
 computation: in particular, it must be a valid `const` (i.e., it cannot
 introduce interior mutability) and it must not require dropping.
 
-For instance, while the previous example was not legal, the following would be:
+For instance the following would be legal even though calls to `do_it` are not
+eligible for implicit promotion:
 
 ```rust
-const BOOL: i32 = {
-  let ret = if cfg!(windows) { 0 } else { 1 };
+const fn do_it(x: i32) -> i32 { 2*x }
+const ANSWER: i32 = {
+  let ret = do_it(21);
   ret
 };
 
-let x: &'static i32 = &BOOL;
+let x: &'static i32 = &ANSWER;
 ```
 
 An access to a `static` is only promotable within the initializer of another
--- a/promotion.md
+++ b/promotion.md
@@ -247,6 +247,21 @@ An access to a `static` is only promotable within the initializer of another
 `static`. This is for the same reason that `const` initializers
 [cannot access statics](const.md#reading-statics).
 
+Crucially, however, the following is *not* legal:
+
+```rust
+const X: Cell<i32> = Cell::new(5); // ok
+const XREF: &Cell<i32> = &X; // not ok
+fn main() {
+    let x: &'static _ = &X; // not ok
+}
+```
+
+Just like allowing `XREF` would be a problem because, by the inlining semantics,
+every user of `XREF` should get their own `Cell`; it would also be a problem to
+promote here because if that code getes executed multiple times (e.g. inside a
+loop), it should get a new `Cell` each time.
+
 ### Named locals
 
 Promotable expressions cannot refer to named locals. This is not a technical
--- a/promotion.md
+++ b/promotion.md
@@ -226,8 +226,8 @@ the result of computing a promoted is a value that does not need dropping.
 
 ### Access to a `const` or `static`
 
-When accessing a `const` in a promotable context, its value anyway gets computed
-at compile-time, so we do not have to check the initializer.  However, the
+When accessing a `const` in a promotable context, its value gets computed
+at compile-time anyway, so we do not have to check the initializer.  However, the
 restrictions described above still apply for the *result* of the promoted
 computation: in particular, it must be a valid `const` (i.e., it cannot
 introduce interior mutability) and it must not require dropping.
@@ -272,12 +272,13 @@ const context, the user likely expects that `x` will live on the stack and be
 initialized at run-time.  Although this is not (to my knowledge) guaranteed by
 the language, we do not wish to violate the user's expectations here.
 
-However, constant-folding still applies: the optimizer may compute `x` at
+Note that constant-folding still applies: the optimizer may compute `x` at
 compile-time and even inline it everywhere if it can show that this does not
 observably alter program behavior.  Promotion is very different from
 constant-folding as promotion can introduce observable differences in behavior
 (if const-evaluation fails) and as it is *guaranteed* to happen in some cases
-(and thus exploited by the borrow checker).
+(and thus exploited by the borrow checker).  This is reflected in the fact that
+promotion affects lifetimes, but constant folding does not.
 
 ### Single assignment
 
--- a/promotion.md
+++ b/promotion.md
@@ -115,56 +115,10 @@ expressions are actually eligible for promotion? We refer to eligible
 expressions as "promotable" and describe the restrictions on such expressions
 below.
 
-### Named locals
-
-Promotable expressions cannot refer to named locals. This is not a technical
-limitation with the CTFE engine. While writing `let x = {expr}` outside of a
-const context, the user likely expects that `x` will live on the stack and be
-initialized at run-time.  Although this is not (to my knowledge) guaranteed by
-the language, we do not wish to violate the user's expectations here.
-
-However, constant-folding still applies: the optimizer may compute `x` at
-compile-time and even inline it everywhere if it can show that this does not
-observably alter program behavior.  Promotion is very different from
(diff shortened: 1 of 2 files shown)


promotion.md
```diff
@@ -137,11 +144,11 @@ resources for little benefit.
 
 ### Access to a `const` or `static`
 
-When accessing a `const` in a promotable context, the restrictions on single
-assignment and named locals do not apply to the body of the `const`. All other
-restrictions, notably that the result of the `const` cannot be `Drop` or mutable
-through a reference still apply. For instance, while the previous example was
```
RalfJung: What are the restrictions about "mutability through a reference here"? I would expect everything listed here to be  restrictions that already come up elsewhere in this document.

ecstatic-morse: That was a less clear way of saying interior mutability. We don't promote `&X` if `X` is a `Cell` for example. I think this should go after the colon along with `Drop` in the revised version.

As an aside, I think we used to not promote accesses to `const`s if the `const` didn't pass [some checks](https://github.com/rust-lang/rust/blob/1.38.0/src/librustc_mir/transform/qualify_consts.rs#L1567), although I think these were related to broken MIR? I was hedging here by mentioning the other restrictions since I still don't understand the reasons for that check. In any case, this will be gone once my final changes to the const-checker land in #66385.

ecstatic-morse:
This looks good to me apart from the mention about promotion of references to types with interior mutability. Perhaps we could include the following as an example?

```rust
const X: Cell<i32> = Cell::new(5); // ok
fn main() {
    let x: &'static _ = &X; // not ok
}
```

promotion.md
RalfJung: A `const` `Cell` should anyway be impossible -- that is a `const`-level restriction, not a promotion-level one. So I am not sure why promotion would need to check anything extra here?

ecstatic-morse: `const X: Cell<i32> = Cell::new(4)` is legal.

RalfJung: > this will be gone once my final changes to the const-checker land in rust-lang/rust#66385.

They have landed, but I am not sure what exactly that means for us here,

RalfJung: > const X: Cell<i32> = Cell::new(4) is legal.

Oh, right I forgot. But `&X` is not legal as the result of a `const` computation.

I think the text I originally wrote said that all the usual const restrictions still apply, did that get lost?

RalfJung: I did some more editing, now it says:

> When accessing a `const` in a promotable context, its value anyway gets computed
at compile-time, so we do not have to check the initializer.  However, the
restrictions described above still apply for the *result* of the promoted
computation: in particular, it must be a valid `const` (i.e., it cannot
introduce interior mutability) and it must not require dropping.

I also moved the "technically not needed" restrictions to the bottom to give them less focus.

But I still wonder what exactly gets checked now that your PR landed, and why checking for interior mutability is not needed any more (if I understood you correctly).

RalfJung:
> This looks good to me apart from the mention about promotion of references to types with interior mutability. Perhaps we could include the following as an example?

Done.

promotion.md
```diff
@@ -259,6 +224,74 @@ or `const` item and refer to that.
...
 
+### Access to a `const` or `static`
+
+When accessing a `const` in a promotable context, its value anyway gets computed
+at compile-time, so we do not have to check the initializer.  However, the
+restrictions described above still apply for the *result* of the promoted
+computation: in particular, it must be a valid `const` (i.e., it cannot
+introduce interior mutability) and it must not require dropping.
+
+For instance, while the previous example was not legal, the following would be:
```
ecstatic-morse: The previous example is now at the end of the document.

RalfJung: Hm. IMO it would be better to use something else as example here, e.g. a `const fn` that is not `rustc_promotable`. That seems less artificial than this. Any proposals?

ecstatic-morse: Nothing that isn't contrived comes to mind. Even just a call to a const fn would be better than the current example IMO.

RalfJung: Done.

promotion.md
RalfJung: Yeah I'm sorry, I went back to my old text and it wasn't there either. Not sure what I was thinking. I added an explicit remark in this PR now.

> The qualifs in the final value of the const get checked, NeedsDrop and HasMutInterior

The drop part makes sense, for interior mutability I somehow expected that would just fall out of the usual rules for const that also have to hold for promoteds... like the example I added showed, this wouldn't even be allowed in explicit consts, let alone in promoteds. But maybe that is exactly what happens?

ecstatic-morse: I see your point. `&Cell::new(4)` is not valid in a `const` context, so it follows that `&CELL_4` is not promotable. I think this case deserves special mention because the value with interior mutability is created in a const context, but the reference is taken in a non-const one, which is subtle to me. I think the example is helpful.

promotion.md
```diff
@@ -259,6 +224,76 @@ or `const` item and refer to that.
 *Dynamic check.* The Miri engine could dynamically check this by ensuring that
 the result of computing a promoted is a value that does not need dropping.
 
+### Access to a `const` or `static`
+
+When accessing a `const` in a promotable context, its value anyway gets computed
+at compile-time, so we do not have to check the initializer.  However, the
+restrictions described above still apply for the *result* of the promoted
+computation: in particular, it must be a valid `const` (i.e., it cannot
+introduce interior mutability) and it must not require dropping.
```
ecstatic-morse: It's worth noting that these restrictions apply *only* in the context lifetime extension. We can initialize an array, which also uses the implicit rules, with a constant containing interior mutability (i.e. `[CELL; 4]`).

promotion.md
RalfJung: > I think this case deserves special mention because the value with interior mutability is created in a const context, but the reference is taken in a non-const one, which is subtle to me.

Agreed. Is what I have enough?

promotion.md
RalfJung: But `[&CELL; 4]` gets rejected, I hope?

promotion.md
ecstatic-morse: Yes.

promotion.md
ecstatic-morse: That compiles on stable because `&T` is `Copy` for all `T`. Any use of it will cause a borrow check error since the access to `CELL` has a temporary lifetime.

promotion.md
```diff
@@ -259,6 +224,76 @@ or `const` item and refer to that.
...
+const X: Cell<i32> = Cell::new(5); // ok
+const XREF: &Cell<i32> = &X; // not ok
+fn main() {
+    let x: &'static _ = &X; // not ok
+}
+```
+
+Just like allowing `XREF` would be a problem because, by the inlining semantics,
+every user of `XREF` should get their own `Cell`; it would also be a problem to
+promote here because if that code getes executed multiple times (e.g. inside a
```
ecstatic-morse:
```suggestion
promote here because if that code gets executed multiple times (e.g. inside a
```

promotion.md
RalfJung: > Promotion fails for &CELL even if it's in an array initializer if that's what you're asking.

Good.

But then behavior seems to be the same for arrays and lifetime extension? It's just that lifetime extension *always* involves a borrow.

promotion.md
```diff
@@ -259,6 +224,76 @@ or `const` item and refer to that.
 *Dynamic check.* The Miri engine could dynamically check this by ensuring that
 the result of computing a promoted is a value that does not need dropping.
 
+### Access to a `const` or `static`
+
+When accessing a `const` in a promotable context, its value anyway gets computed
```
oli-obk: I think the `anyway` belongs after `compile-time`

promotion.md
```diff
@@ -259,6 +224,76 @@ or `const` item and refer to that.
...
+Promotable expressions cannot refer to named locals. This is not a technical
+limitation with the CTFE engine. While writing `let x = {expr}` outside of a
+const context, the user likely expects that `x` will live on the stack and be
+initialized at run-time.  Although this is not (to my knowledge) guaranteed by
+the language, we do not wish to violate the user's expectations here.
+
+However, constant-folding still applies: the optimizer may compute `x` at
+compile-time and even inline it everywhere if it can show that this does not
+observably alter program behavior.  Promotion is very different from
+constant-folding as promotion can introduce observable differences in behavior
```
oli-obk: I think one important think to note here is that const prop can never cause you to get a `'static` lifetime if you wouldn't have gotten it without const prop. Const prop can only cause lints and performance optimizations

RalfJung: Like this?

RalfJung:
@ecstatic-morse any more concerns/comments?

ecstatic-morse:
r=me

PR #33, opened 2019-12-26
expand on the Send/Sync situation for consts

RalfJung:
I think conceptually it makes more sense to start from "`const` ought to be `Send`" and then argue that thus, the data behind shared references should be `Sync`.

### DIFF
--- a/const.md
+++ b/const.md
@@ -80,12 +80,14 @@ ecosystem that would break if we just started enforcing this now. See
 [this issue](https://github.com/rust-lang/rust/issues/49206) and the
 [PR attempting to fix this](https://github.com/rust-lang/rust/pull/54424/).
 
-Also, one could make the argument that the value does not have to be `Send`
-because it is not actually sent to other threads; instead, conceptually, each
-thread re-does the same computation.  But we know they will all come to the same
+One could make the argument that the value does not have to be `Send` because it
+is not actually sent to other threads; instead, conceptually, each thread
+re-does the same computation.  But we know they will all come to the same
 result.  This works, except when we consider address identity: with references
 in the `const`, all threads will get the same address, unlike in case of a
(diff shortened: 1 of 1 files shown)


const.md
```diff
@@ -67,18 +67,28 @@ fine: inlining that reference everywhere has the same behavior as computing a
...
+One could make the argument that the value does not have to be `Send` because it
+is not actually sent to other threads; instead, conceptually, each thread
+re-does the same computation.  But we know they will all come to the same
+result.  This works, except when we consider address identity: with references
+in the `const`, all threads will get the same address, unlike in case of a
+per-thread recomputation which would lead to different addresses. As a
+consequence, non-`Send` `const` without references are fine, but once references
+and thus address identity comes into play, we have a problem.
+
 *Dynamic check.* It is unclear how the Miri engine could dynamically check this.
```
oli-obk: I started implementing this before we had validation because I thought we could scrap some of const qualif for a value analysis. What I was basically doing is what validation is doing, but looked for explicit `unsafe impl Send` or `unsafe impl Sync` on the type of the value currently being validated. Aggregates that got their trait impls due to the auto trait nature and not explicit impls were simply walked. This permitted `Option<*const u8>::None` but forbade `Option<*const u8>::Some(&1)`

RalfJung: I doubt we can actually get away with forbidding `Option<*const u8>::Some(&1)` though...

> What I was basically doing is what validation is doing, but looked for explicit unsafe impl Send or unsafe impl Sync on the type of the value currently being validated.

Yeah, I had thoughts along those lines, too. Unfortunately though, negative impls aren't stable yet, so it seems plausible people might do things like add a `PhantomData<*const ()>` field as a "no-`Sync`" marker... :/

PR #34, opened 2019-12-26
update status of dynamic checks

RalfJung:
Update docs to the recent dynamic check changes by @oli-obk. Cc @ecstatic-morse 

Also, I think there's no sensible dynamic checks that can be done for promoteds, right? They have to be valid consts, of course, but beyond that there's nothing I think.

### DIFF
--- a/promotion.md
+++ b/promotion.md
@@ -120,7 +120,7 @@ restrictions described there are needed because we want `const` to behave the
 same as copying the `const` initializer everywhere the constant is used; we need
 the same property when promoting expressions. But we need more.
 
-Note that there is no point in doing dynamic checks here.  The entire point of
+Note that there is no point in doing additional dynamic checks here.  The entire point of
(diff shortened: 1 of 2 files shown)


promotion.md
```diff
@@ -120,6 +120,13 @@ restrictions described there are needed because we want `const` to behave the
 same as copying the `const` initializer everywhere the constant is used; we need
 the same property when promoting expressions. But we need more.
 
+Note that there is no point in doing dynamic checks here.  The entire point of
```
oli-obk:
```suggestion
Note that there is no point in doing additional dynamic checks here.  The entire point of
```

oli-obk:
> Also, I think there's no sensible dynamic checks that can be done for promoteds, right? They have to be valid consts, of course, but beyond that there's nothing I think.

Yea, the final value of a promoted just needs the same checks applied that the final value of a constant must get.

PR #40, opened 2020-04-11
describe static vs dynamic const checks

RalfJung:
Basically putting https://github.com/rust-lang/const-eval/issues/11#issuecomment-612376089 into the repo as I figured it could be useful.

### DIFF
--- a/const_checks.md
+++ b/const_checks.md
@@ -1,0 +1,29 @@
+# Static and dynamic checks
+
+This repository describes a set of rules that various forms of compile-time evaluated code needs to satisfy.
+The compiler contains two kinds of checks that guard against violations of these rules: static checks and dynamic checks.
+
+## Dynamic checks
+
+Dynamic checks are conceptually very simple: when evaluating the compile-time code, we look at what it does, and if what it does violates the rules, we halt evaluation.
+Thus, a dynamic check generally makes it very clear what is being protected against.
+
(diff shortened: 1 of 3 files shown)


const_checks.md
```diff
@@ -0,0 +1,29 @@
...
+This repository describes a set of rules that various forms of compile-time evaluated code needs to satisfy.
+The compiler contains two kinds of checks that guard against violations of these rules: static checks and dynamic checks.
+
+## Dynamic checks
+
+Dynamic checks are conceptually very simple: when evaluating the compile-time code, we look at what it does, and if what it does violates the rules, we halt evaluation.
+Thus, a dynamic check generally makes it very clear what is being protected against.
+
+The main disadvantage of dynamic checks is that they can only run when the compile-time code is being evaluated, which is after monomorphization.
+We generally try to avoid post-monomorphization errors as they make for a bad user experience.
```
oli-obk, line 12:
```suggestion
We generally try to avoid post-monomorphization errors as they inherently make for a bad user experience.
While there are technical aspects that could be improved here, the main problem is that the site where the error is reported is disconnected from the site where the root cause is.
Such problems can be observed when creating an associated constant that uses associated constants from generic parameters.
These generic parameters are unknown, so the usage of these associated constants may cause errors depending on the *value* of the generic parameter's associated constants.
```

PR #48, opened 2020-07-28
add document on consts in patterns

RalfJung:
Write down what I know about consts in patterns. There are some TODO here that I'll need help to fill in -- Cc @oli-obk @lcnr.

Cc https://github.com/rust-lang/const-eval/issues/42

### DIFF
--- a/patterns.md
+++ b/patterns.md
@@ -1,0 +1,62 @@
+# Constants in patterns
+
+Constants that pass the [structural equality check](https://github.com/rust-lang/rust/issues/74446) can be used in patterns:
+```rust
+const FOO: i32 = 13;
+
+fn is_foo(x: i32) -> bool {
+  match x {
+    FOO => true,
+    _ => false,
+  }
+}
+```
+However, that check has some loopholes, so e.g. `&T` can be used in a pattern no matter the `T`.
+
+A const-pattern is compiled by calling `PartialEq::eq` to compare the subject of the `match` with the constant,
+except for TODO where the constant is treated as if it was inlined as a pattern (and the usual `match` tree is constructed).
+
+## Soundness concerns
+
+Most of the time there are no extra soundness concerns due to const-patterns; except for surprising behavior nothing can go wrong when the structural equality check gives a wrong answer.
+However, there is one exception: some constants participate in exhaustiveness checking.
+If a pattern is incorrectly considered exhaustive, that leads to a critical soundness bug.
+
+Exhaustiveness checking is done for constants of type TODO, as well as `&[u8]` and `&str` (but no other types with references).
+This means we can write:
+```rust
+#![feature(exclusive_range_pattern)]
+#![feature(half_open_range_patterns)]
+const PAT: &[u8] = &[0];
+
+pub fn test(x: &[u8]) -> bool {
+    match x {
+        PAT => true,
+        &[] => false,
+        &[1..] => false,
+        &[_, _, ..] => false
+    }
+}
+```
+
+To ensure soundness of exhaustiveness checking, it is crucial that all data considered this check is fully immutable.
+In particular, for constants of reference type, it is important that they only point to immutable data.
+For this reason, the static const checks reject references to `static` items.
+This is a new soundness concern that otherwise does not come up during CTFE (where *reading from* a mutable `static` during const initialization is a problem, but just referencing a mutable `static` is not).
+A more precise check could be possible, but is non-trivial: even an immutable `static` could point to a mutable `static`; that would have to be excluded.
+(We could, in principle, also rely on it being UB to have a shared reference to mutable memory, but for now we prefer not to rely on the aliasing model like that---aliasing of global pointers is a tricky subject.)
+
+*Dynamic check.*
+To dynamically ensure constants only point to read-only data, we maintain a global invariant:
+pointers to a `static` (including memory that was created as part of interning a `static`) may never "leak" to a non-`static`.
+All memory outside of `static`s is marked as immutable.
+As a consequence, non-`static`s recursively only point to immutable memory.
+
+This check is uncomfortably close to the static checks, and fragile due to its reliance on a global invariant.
+Longer-term, it would be better to move to a more local check, by making validation check if consts are recursively read-only.
+This check (unlike our current validation) would have to descend through pointers to `static`.
+
+An alternative could be to wait for the [value tree proposal](https://github.com/rust-lang/compiler-team/issues/323) to be implemented.
+Then we could check during value tree construction that all memory is read-only (if we use the CTFE machine in `const`-mode, that will happen implicitly), and thus we know for sure we can rely on the value tree to be sound to use for exhaustiveness checking.
+How exactly this looks like depends on how we resolve [pattern matching allowing many things that do not fit a value tree](https://github.com/rust-lang/rust/issues/74446#issuecomment-663439899).
+However, if only the well-behaved part of the value tree is used for exhaustiveness checking, handling these extra patterns should not interact closely with the soundness concerns discussed here.
--- a/patterns.md
+++ b/patterns.md
@@ -25,7 +25,7 @@ Most of the time there are no extra soundness concerns due to const-patterns; ex
 However, there is one exception: some constants participate in exhaustiveness checking.
 If a pattern is incorrectly considered exhaustive, that leads to a critical soundness bug.
 
-Exhaustiveness checking is done for constants of type TODO, as well as `&[u8]` and `&str` (but no other types with references).
+Exhaustiveness checking is done for all constants that do not have reference type, as well as `&[u8]` and `&str`.
 This means we can write:
 ```rust
 #![feature(exclusive_range_pattern)]
--- a/patterns.md
+++ b/patterns.md
@@ -13,8 +13,7 @@ fn is_foo(x: i32) -> bool {
 ```
 However, that check has some loopholes, so e.g. `&T` can be used in a pattern no matter the `T`.
 
-A const-pattern is compiled by [calling `PartialEq::eq`][compile-partial-eq] to compare the subject of the `match` with the constant,
-except for TODO where the constant is treated as if it was inlined as a pattern (and the usual `match` tree is constructed).
+Any reference type const-pattern is compiled by [calling `PartialEq::eq`][compile-partial-eq] to compare the subject of the `match` with the constant. Const-patterns with other types (enum, struct, tuple, array) are treated as if the constant was inlined as a pattern (and the usual `match` tree is constructed).
 
 [struct-eq]: https://github.com/rust-lang/rust/blob/2c28244cf0fc9868f55070e55b8f332d196eaf3f/src/librustc_mir_build/hair/pattern/const_to_pat.rs#L121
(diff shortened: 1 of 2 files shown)


patterns.md
```diff
@@ -0,0 +1,62 @@
...
+pointers to a `static` (including memory that was created as part of interning a `static`) may never "leak" to a non-`static`.
+All memory outside of `static`s is marked as immutable.
+As a consequence, non-`static`s recursively only point to immutable memory.
+
+This check is uncomfortably close to the static checks, and fragile due to its reliance on a global invariant.
+Longer-term, it would be better to move to a more local check, by making validation check if consts are recursively read-only.
+This check (unlike our current validation) would have to descend through pointers to `static`.
+
+An alternative could be to wait for the [value tree proposal](https://github.com/rust-lang/compiler-team/issues/323) to be implemented.
+Then we could check during value tree construction that all memory is read-only (if we use the CTFE machine in `const`-mode, that will happen implicitly), and thus we know for sure we can rely on the value tree to be sound to use for exhaustiveness checking.
```
RalfJung, line 60: @oli-obk the more I think about it, the more I like this: as far as the dynamic checks are concerned, constants can point to mutable data just fine as long as we never construct a value tree for them (or as long as the mutable part is in the "fallback" part of the value tree that we might have to support "bad" patterns that do not fit a value tree, like raw pointers, function pointers, floats).

oli-obk: The problem isn't just a value tree pointing to a mutable allocation, it's a mutable allocation being used during the const evaluation of the initializer of a constant that will be converted to a value tree. It's not really unsound and I don't know of any problems, we just have had the policy of "running the initializers of a constant manually (copy paste) gives the same result as reading the constant", and this would break it.

RalfJung: That is an entirely orthogonal concern (not even specific to patterns) that we [already cover here](https://github.com/rust-lang/const-eval/blob/master/const.md#reading-statics). But I think you are mixing up *initializing a const reads from a static that could later be mutated* and *the final value of a const (transitively) points to a mutable static*. These are two very different and orthogonal concerns, in my view.

patterns.md
```diff
@@ -0,0 +1,62 @@
...
+A const-pattern is compiled by calling `PartialEq::eq` to compare the subject of the `match` with the constant,
+except for TODO where the constant is treated as if it was inlined as a pattern (and the usual `match` tree is constructed).
+
+## Soundness concerns
+
+Most of the time there are no extra soundness concerns due to const-patterns; except for surprising behavior nothing can go wrong when the structural equality check gives a wrong answer.
+However, there is one exception: some constants participate in exhaustiveness checking.
+If a pattern is incorrectly considered exhaustive, that leads to a critical soundness bug.
+
+Exhaustiveness checking is done for constants of type TODO, as well as `&[u8]` and `&str` (but no other types with references).
```
oli-obk, line 25: Exhaustiveness checks are done for constants of all types except references and with references only for `&[u8]` and `&str`. I don't know how to possibly exclude these constants as they also have a literal version `b"foo"` and `"foo"`, which are indistinguishable at the pattern checking stage. Even if we somehow set it up that named constants don't get evaluated until necessary, it feels very fragile to require lazy evaluation for such a semantic information.

patterns.md
oli-obk: ... I had something written and then realized we are only talking about dynamic checks... Yea, it's all fine for dynamic checks. I just can't come up with static checks that are more relaxed than today's.

patterns.md
RalfJung: *All* types? I hope not for raw pointers and function pointers, I cannot imagine that to be soundly possible...

RalfJung: If you have a link to the relevant code, that might be good to explain (AFAIK there is no RFC we could link to).

patterns.md
RalfJung: Right, I see static checks as something we can only really start talking about once we have established what the dynamic property is that we want to approximate.

patterns.md
oli-obk: There are several points:

* most of the stuff that takes apart constants is https://github.com/rust-lang/rust/blob/2c28244cf0fc9868f55070e55b8f332d196eaf3f/src/librustc_mir_build/hair/pattern/const_to_pat.rs#L204 This is what makes constants indistinguishable from writing the patterns directly
* structural match checks happen here: https://github.com/rust-lang/rust/blob/2c28244cf0fc9868f55070e55b8f332d196eaf3f/src/librustc_mir_build/hair/pattern/const_to_pat.rs#L121
* actual reference comparisons via `PartialEq::eq` happen [here](https://github.com/rust-lang/rust/blob/2c28244cf0fc9868f55070e55b8f332d196eaf3f/src/librustc_mir_build/build/matches/test.rs#L355) but this is a red herring, that's just codegen, even though lots of failures happen there if the previous logic mucks up
* some lazy dereferencing happens [here](https://github.com/rust-lang/rust/blob/2c28244cf0fc9868f55070e55b8f332d196eaf3f/src/librustc_mir_build/hair/pattern/_match.rs#L321) in the middle of match logic. Likely this is also a red herring, but I fixed some ICEs there before fully understanding what I was doing
* checking slice patterns against constants of slice type is done [here](https://github.com/rust-lang/rust/blob/2c28244cf0fc9868f55070e55b8f332d196eaf3f/src/librustc_mir_build/hair/pattern/_match.rs#L2123)

RalfJung: Hm, none of that seems to have to do anything with exhaustiveness checking?

patterns.md
```diff
@@ -0,0 +1,63 @@
...
+fn is_foo(x: i32) -> bool {
+  match x {
+    FOO => true,
+    _ => false,
+  }
+}
+```
+However, that check has some loopholes, so e.g. `&T` can be used in a pattern no matter the `T`.
+
+A const-pattern is compiled by calling `PartialEq::eq` to compare the subject of the `match` with the constant,
```
RalfJung: Is it really always `PartialEq`? I would imagine that e.g.
```rust
const TEST: (i32, i32) = (0, 0);

match (x, y) {
    (z1, z2) if z1 != z2 => ...,
    TEST => ....
    (z, _) => ...
}
```
compiles the same as when `TEST` is inlined?

patterns.md
oli-obk: Well, they all have effects on exhaustiveness (because they change the datastructures that exhaustiveness works on) or are invoked by exhaustiveness. The actual exhaustivness entry points are https://github.com/rust-lang/rust/blob/2c28244cf0fc9868f55070e55b8f332d196eaf3f/src/librustc_mir_build/hair/pattern/_match.rs#L823 and https://github.com/rust-lang/rust/blob/2c28244cf0fc9868f55070e55b8f332d196eaf3f/src/librustc_mir_build/hair/pattern/_match.rs#L974 and https://github.com/rust-lang/rust/blob/2c28244cf0fc9868f55070e55b8f332d196eaf3f/src/librustc_mir_build/hair/pattern/_match.rs#L2069 (the latter being where constants and slices and arrays are hardcoded)

patterns.md
oli-obk: We are taking it apart, but we should not. All constants should be treated as opaque according to the status quo of what has been RFCed

patterns.md
RalfJung: Okay somehow I expected a `match` where I'd clearly see references being treated separately. I guess I'll just transcribe what you said into the text, there is no way for me to verify that.

patterns.md
RalfJung: @oli-obk  could you propose something concrete we could write for the remaining TODO?

patterns.md
RalfJung: I adjusted the text, please check. :)

patterns.md
```diff
@@ -0,0 +1,66 @@
...
+  match x {
+    FOO => true,
+    _ => false,
+  }
+}
+```
+However, that check has some loopholes, so e.g. `&T` can be used in a pattern no matter the `T`.
+
+A const-pattern is compiled by [calling `PartialEq::eq`][compile-partial-eq] to compare the subject of the `match` with the constant,
+except for TODO where the constant is treated as if it was inlined as a pattern (and the usual `match` tree is constructed).
```
oli-obk, line 17:
```suggestion
Any reference type const-pattern is compiled by [calling `PartialEq::eq`][compile-partial-eq] to compare the subject of the `match` with the constant. Const-patterns with other types (enum, struct, tuple, array) are treated as if the constant was inlined as a pattern (and the usual `match` tree is constructed).
```

patterns.md
oli-obk: I wrote something up. Should we also mention that this diverges from the RFCed semantics?

patterns.md
```diff
@@ -0,0 +1,66 @@
...
+  match x {
+    FOO => true,
+    _ => false,
+  }
+}
+```
+However, that check has some loopholes, so e.g. `&T` can be used in a pattern no matter the `T`.
+
+Any reference type const-pattern is compiled by [calling `PartialEq::eq`][compile-partial-eq] to compare the subject of the `match` with the constant. Const-patterns with other types (enum, struct, tuple, array) are treated as if the constant was inlined as a pattern (and the usual `match` tree is constructed).
+Note that this is in contradiction with what has been RFC'd, which specifies that we should always call `PartialEq::eq`.
```
RalfJung, line 17: Would be good to link to the RFC here. I assume that's https://github.com/rust-lang/rfcs/blob/master/text/1445-restrict-constants-in-patterns.md but so far I have not found where it says that we should use `PartialEq`.

patterns.md
oli-obk: :laughing: ok you had the same thought^^

patterns.md
RalfJung: The RFC calls this "semantic equality" and makes some preparation for maybe switching to it in the future, but it dosn't actually say that we want to adopt semantic equality. That's my reading, anyway.

oli-obk: It's a bit spread around, but https://github.com/rust-lang/rfcs/blob/master/text/1445-restrict-constants-in-patterns.md#disadvantage-weakened-abstraction-bounary is probably the best entry point

RalfJung: But later the RFC explicitly says

> The goal of this RFC is not to decide between semantic and structural equality. Rather, the goal is to restrict pattern matching to that subset of types where the two variants behave roughly the same.

oli-obk: It's very explicit in the [detailed design section](https://github.com/rust-lang/rfcs/blob/master/text/1445-restrict-constants-in-patterns.md#the-structural-match-attribute)

>  When converting a constant value into a pattern, if the constant is of struct or enum type, we will check whether this attribute is present on the struct -- if so, we will convert the value as we do today. If not, we will report an error that the struct/enum value cannot be used in a pattern.

where "as we do today" is "convert to an equivalent pattern"

RalfJung: I don't think that is more explicit that "The goal of this RFC is not to decide between semantic and structural equality".

That text you quoted does not say anywhere that we should use semantic equality for anything.

RalfJung: I adjusted the text to reflect what I think the RFC says.

RalfJung:
Thanks for the help @oli-obk :)

PR #50, opened 2020-08-07
Skill tree additions and restructurings

oli-obk:


src/skill_tree.md
```diff
@@ -75,10 +76,38 @@ items = []
...
+name = "offset_of"
+label = "offset_of"
+items = [
+  { label = "offset_from", href = "https://github.com/rust-lang/rust/issues/41079", port = "offset_from" },
+]
+href = "https://github.com/rust-lang/rust/issues/71499"
+requires = [
+  "unconst_rules:raw_ptr_deref",
+  "raw_ref_macros",
+  "maybe_uninit_as_ptr",
```
RalfJung, line 95: `const_offset_from` is missing.

RalfJung:
Other things:
* the "mutable ref in final value" node needs renaming; the feature we want is "mutable ref in `const` items".
* `<int>::from_str` as a specific subitem of `FromStr` seems rather arbitrary? I'd probably remove it, that seems less confusing.
* would it make sense to give the feature name everywhere? That makes it easier to match things up. (I have been adding the feature name to issue titles wherever I noticed they are missing, which makes the tracing issues much easier to find in my browser history.)

src/skill_tree.md
```diff
@@ -75,10 +76,38 @@ items = []
 name = "unconst_rules"
 label = "Need to come up\nwith a scheme\nfor doing unsafe\nin const fn"
 items = [
-  { label = "transmute", href = "https://github.com/rust-lang/rust/issues/53605" }
+  { label = "transmute", href = "https://github.com/rust-lang/rust/issues/53605" },
```
RalfJung, line 79:
```suggestion
  { label = "transmute in functions", href = "https://github.com/rust-lang/rust/issues/53605" },
```

src/skill_tree.md
```diff
@@ -75,10 +76,38 @@ items = []
...
+  { label = "union field access", href = "https://github.com/rust-lang/rust/issues/51909", port = "union" },
```
RalfJung, line 80:
```suggestion
  { label = "union field access in functions", href = "https://github.com/rust-lang/rust/issues/51909", port = "union" },
```

src/skill_tree.md
```diff
@@ -75,10 +76,38 @@ items = []
 name = "unconst_rules"
 label = "Need to come up\nwith a scheme\nfor doing unsafe\nin const fn"
```
RalfJung, line 77:
```suggestion
label = "Need to come up\nwith a scheme\nfor const unsafe/unconst"
```

oli-obk:
>     * `<int>::from_str` as a specific subitem of `FromStr` seems rather arbitrary? I'd probably remove it, that seems less confusing.

that has its own issue, thus I added it at all.

>     * would it make sense to give the feature name everywhere?

yes, but that makes everything more verbose. I've been trimming down the graph where I can

src/skill_tree.md
```diff
@@ -119,7 +123,12 @@ name = "mutex_new"
 label = "Mutex::new"
 href = "https://github.com/rust-lang/rust/issues/66806"
 items = []
-requires = ["parking_lot", "unconst_rules"]
```
RalfJung: Shouldn't `Mutex::new` still require `parking_lot`?

RalfJung:
>  that has its own issue, thus I added it at all.

`FromStr` has a tracking issue. But I was talking about the `<int>::from_str`.

> yes, but that makes everything more verbose. I've been trimming down the graph where I can

I mean it makes it more informative, I wouldn't call that "verbose".^^

oli-obk:
I added all the feature gate names where on existed

RalfJung:
Awesome!

What does `fg:` mean?

oli-obk:
I moved to `feature` and did some smaller textual changes that shrink the tree's width again

RalfJung:
Thanks a lot! Let's merge this so we can see how it looks rendered.

PR #52, opened 2020-08-10
Link to skill tree and fix some nodes

RalfJung:
r? @oli-obk 
Fixes https://github.com/rust-lang/const-eval/issues/51

### DIFF
--- a/src/skill_tree.md
+++ b/src/skill_tree.md
@@ -137,7 +137,7 @@ items = []
 name = "const_fn_in_patterns"
 label = "const fn callable in patterns"
 href = "https://github.com/rust-lang/rust/issues/57240"
-requires = ["rfc:2920"]
+requires = ["const-blocks"]
 items = []
 
 [[group]]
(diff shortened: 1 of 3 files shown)


src/skill_tree.md
```diff
@@ -84,8 +84,7 @@ href = "https://github.com/rust-lang/const-eval/issues/14"
 
 [[group]]
 name = "offset_of"
-label = "offset_of\nfeature:const_ptr_offset"
-href = "https://github.com/rust-lang/rust/issues/71499"
```
RalfJung, line 88: `offset_of` has very little to do with `offset`/`wrapping_offset`. If we want the latter in the skill tree, they should be a separate node.

src/skill_tree.md
```diff
@@ -179,10 +184,9 @@ href = "https://github.com/rust-lang/rust/issues/74925"
 items = []
 
 [[group]]
-name = "rfc"
```
oli-obk, line 182: this used to contain more RFCs, but they have been moved to tracking issues since. I was thinking that we should somehow highlight that some node is an RFC and not something already accepted (though... considering the random collection of issues I opened that doesn't have a merged RFC, I guess we can just treat them the same)

RalfJung:
Hm, https://rust-lang.github.io/const-eval/ still contains that "rfc" node?

RalfJung:
CI said
```
Warning: The mdbook-skill-tree preprocessor was built against version 0.3.7 of mdbook, but we're being called from version 0.4.1
thread 'main' panicked at 'encountered error missing field `items` for key `group` at line 279 column 1', mdbook-skill-tree/src/preprocessor.rs:85:31
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2020-08-10 13:22:09 [ERROR] (mdbook::utils): Error: The preprocessor exited unsuccessfully
##[error]Process completed with exit code 101.
```

RalfJung:
I did a direct push to hopefully fix this.

PR #54, opened 2020-08-13
update promotion docs

RalfJung:
The promotion documentation was partially outdated (like saying that in const initializers, we wouldn't actually promote) and partially confusing (like the `Vec::new` example). I hope this helps.

Cc @rust-lang/wg-const-eval

### DIFF
--- a/promotion.md
+++ b/promotion.md
@@ -1,8 +1,8 @@
 # Const promotion
 
-"Promotion" is the act of guaranteeing that code *not* written in an (explicit)
-const context will be run at compile-time. Explicit const contexts include the
-initializer of a `const` or `static`, or an array length expression.
+"Promotion" is the act of splicing a part of a MIR computation out into a
+separate self-contained MIR body which is evaluated at compile-time like a
+constant.
 
 ## Promotion contexts
 
@@ -35,7 +35,7 @@ non-`Copy` types to be initialized idiomatically, for example
 
 [RFC 2203]: https://github.com/rust-lang/rfcs/blob/master/text/2203-const-repeat-expr.md
 
-### `#[rustc_args_required_const(...)]`
+### `#[rustc_args_required_const(...)]` and inline assembly `const` operands
 
 Additionally, some platform intrinsics require certain parameters to be
 immediates (known at compile-time). We use the `#[rustc_args_required_const]`
@@ -44,59 +44,46 @@ attribute, introduced in
 specify these parameters and (aggressively, see below) try to promote the
 corresponding arguments.
 
-### Implicit and explicit contexts
+Similarly, inline assembly has `const` operands, which are treated the same way
+as `rustc_args_required_const` arguments.
+
+## Implicit and explicit promotion
 
 On top of what applies to [consts](const.md), promoteds suffer from the additional issue that *the user did not ask for them to be evaluated at compile-time*.
 Thus, if CTFE fails but the code would have worked fine at run-time, we broke the user's code for no good reason.
 Even if we are sure we found an error in the user's code, we are only allowed to [emit a warning, not a hard error][warn-rfc].
-That's why we have to be very conservative with what can and cannot be promoted.
-
-For example, users might be surprised to learn that whenever they take a
-reference to a temporary, that temporary may be promoted away and never
-actually put on the stack. In this way, lifetime extension is an "implicit
-promotion context": the user did not ask for the value to be promoted.
-
-On the other hand, when a user passes an expression to a function with
-`#[rustc_args_required_const]`, the only way for this code to compile is to promote it.
-In that sense, the user is explicitly asking for that expression
-to be evaluated at compile-time even though they have not written it in a
-`const` declaration. We call this an "explicit promotion context".
-
-Currently, non-`Copy` array initialization is treated as an implicit context,
-because the code could compile even without promotion (namely, if the result
-type is `Copy`).
+We call this *implicit* promotion, and we have to be very conservative with what can and cannot be implicitly promoted.
 
 CTFE of implicitly promoted code must never fail to evaluate except if the
 run-time code also would have failed. This means we cannot permit calling
-arbitrary `const fn`, as we cannot predict if they are going to perform an
-["unconst" operation](const_safety.md). Thus, only functions marked
-`#[rustc_promotable]` are implicitly promotable (see below). See
-[rust-lang/const-eval#19](https://github.com/rust-lang/const-eval/issues/19) for
-a thorough discussion of this. At present, this is the only difference between
-implicit and explicit contexts. The requirements for promotion in an implicit
-context are a superset of the ones in an explicit context.
+arbitrary `const fn`, as discussed in detail in
+[rust-lang/const-eval#19](https://github.com/rust-lang/const-eval/issues/19).
+Thus, only functions marked `#[rustc_promotable]` are implicitly promotable (see
+below).
+
+On the other hand, when a user passes an expression to a function with
+`#[rustc_args_required_const]`, the only way for this code to compile is to
+promote it.  In that sense, the user is explicitly asking for that expression to
+be evaluated at compile-time even though they have not written it in a `const`
+declaration. We can thus be less conservative. This is called *explicit*
+promotion.
+
+Currently, the following are considered explicit promotion contexts:
+* `#[rustc_args_required_const]` arguments and inline assembly `const` operands everywhere.
+* Everything inside the bodies of `const` and `static` items. (Note: this is handled separately from "explicit contexts" in promotion analysis, but the effect is the same.)
+
+In these contexts, we promote calls to arbitrary `const fn`.
+
+There is one further special case for the bodies of `const` and `static` items; here we additionally promote union field accesses.
+Both of these special cases can lead to promoting things that can fail to evaluate.
+Currently, this works out because it just leads to a warning, but longer-term it would be desirable to turn evaluation failures into hard errors, which for these promoteds means we have to guarantee that we only evaluate them on-demand.
+
+[See below][static access] for another special case in promotion analysis:
+accesses and references to statics are only promoted inside other statics.
 
 [warn-rfc]: https://github.com/rust-lang/rfcs/blob/master/text/1229-compile-time-asserts.md
 
-### Promotion contexts inside `const` and `static`
-
-Lifetime extension is also responsible for making code like this work:
-
-```rust
-const FOO: &'static i32 = {
-    let x = &13;
-    x
-};
-```
-
-We defined above that promotion guarantees that code in a non-const context
-will be executed at compile-time. The above example illustrates that lifetime
-extension and non-`Copy` array initialization are useful features *inside*
-`const`s and `static`s as well. Strictly speaking, the transformation used to
-enable these features inside a const-context is not promotion; no `promoted`s
-are created in the MIR.  However the same rules for promotability are used with
-one modification: Because the user has already requested that this code run at
-compile time, all contexts are treated as explicit.
+## "enclosing scope" rule
 
 Notice that some code involving `&` *looks* like it relies on lifetime
 extension but actually does not:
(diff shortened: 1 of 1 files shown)


promotion.md
```diff
@@ -105,12 +68,45 @@ extension but actually does not:
...
+promote it.  In that sense, the user is explicitly asking for that expression to
+be evaluated at compile-time even though they have not written it in a `const`
+declaration. We can thus be less conservative. This is called *explicit*
+promotion.
+
+Currently, the following are considered explicit promotion contexts:
+* `#[rustc_args_required_const]` arguments
+* lifetime extension in the bodies of `const` and `static` items and array lengths
+
+Everything else is an implicit promotion context, including `const fn` bodies and non-`Copy` array initializers.
```
oli-obk, line 102: `const fn` bodies are also explicit promotion contexts:

```rust
const fn foo() -> i32 {
    42
}

const fn bar() -> &'static i32 {
    &foo()
}
```

RalfJung: But... that's bad? They can be evaluated at runtime!

oli-obk: sure, but you can't do anything bad as the expression must already be const evaluable

RalfJung: You can do tons of bad things, like divide by 0 or panic. This breaks everything, it means we cannot make const-eval failures hard errors.  Just imagine doing codegen for
```rust
fn evil() { /* do something nasty */ }

fn bar() {
  if false { &evil() }
}
```
We have to evaluate the promoted for codegen, but if evaluation fails we cannot cause a hard error as that would mean we broke the build for no good reason.

Since control flow has just gotten stable in const, can we quickly fix this hole and make `const fn` *implicit* promotion contexts?

oli-obk: > Since control flow has just gotten stable in const, can we quickly fix this hole and make `const fn` _implicit_ promotion contexts?

turning it into implicit promotion contexts is a breaking change that is entirely independent of control flow

RalfJung: Sure, but without control flow there's no dead code.

... not sure how that helps mitigate the scope of the breaking change though. But I feel like this is still something we should try ASAP, as otherwise all hope is lost of ever treating promoteds in a sane way. :(

RalfJung: This is what happens when we do not write down proper design documents. :( I "naturally" assumed that `const fn` are treated like `fn` for promotion (because they can be evaluated at runtime, and thus all of the problems of promotion in `fn` also apply in `const fn`), and you "naturally" assumed that `const fn` are treated like `const`/`static` initializers. And since I cannot actually understand the code that implements this and we never wrote any of this down at a high level, here we are...

ecstatic-morse: Oh no. I wish I would have realized this while doing the initial refactoring of const-checking. That logic is [here](https://github.com/rust-lang/rust/blame/master/src/librustc_mir/transform/promote_consts.rs#L761-L771) BTW.

RalfJung: I wish I had read the code more careful to be able to provide such input more timely. No point lamenting over spilled beans, though I wonder if there is something we can do with our process to help with such things, without drowning in paperwork. The MCP mechanism could have helped, assuming that such details of the desired analysis would be spelled out in the MCP.

>  That logic is here BTW

This looks like `const fn` are not an explicit promotion context, but get some kind of special treatment that is half-way between explicit and implicit contexts? Are there other such "hybrid" cases?

> I can do a crater run to see how many people depend on this.

Yes that would be great.

ecstatic-morse: Note to self: this has worked for every stable release after 1.35.

RalfJung:
When looking at the code, I found something else this document didn't mention yet: promotion of inline assembly arguments. In the future, when landing such things, could you please open an issue to adjust the docs, or (even better) a PR?

Is there anything else that I missed?

RalfJung:
Here are some more strange things I found -- differences in promotion behavior not captured by the implicit/explicit split, and not documented here:
* There's a `const_kind` field of type `Option<ConstContext>` that is undocumented unfortunately, so I have no idea what `None` means. Looks like this is filled with `body_const_context`, so `None` means non-`const` `fn`? This seems to be the thing that introduces some extra differences.
* Inside a `static mut`, `&mut [...]` is also promoted. Why does that make any sense? Promoteds are consts I thought, they have read-only memory, don't they?
* There's a comment saying `&mut []` is promoted only inside functions, but if I understand the code correctly this is actually just done inside non-`const` `fn`.
* There's a special [here](https://github.com/rust-lang/rust/blob/860d599d5b11db275a829a3f898560405fa84bbe/src/librustc_mir/transform/promote_consts.rs#L523) where I cannot figure out what it does.
* union field accesses are checked depending on the `const_kind` (not explicit vs implicit as I expected).
* I don't understand [this one](https://github.com/rust-lang/rust/blob/860d599d5b11db275a829a3f898560405fa84bbe/src/librustc_mir/transform/promote_consts.rs#L584) at all. Consts cannot refer to statics, sure, but promotion isn't the place to *enforce* that property, or is it?
* `validate_rvalue` contains two more `const_kind.is_none()`.

So to conclude, we have an entire suite of differences that has nothing to do with explicit vs implicit, and that was never mentioned in any discussion or documented anywhere :fearful: . This mess has dimensions I was not aware of, and I was paying close attention on this topic for years now. Something is very wrong with our approach here.

And I still don't know how this allows arbitrary promotion in `const fn`.

RalfJung:
I just hope this is not representative of the qualttiy of our design documents for other things like const qualification or pattern matching. If I have to read all that code to verify that the documents make sense, then what is the point of having such documents? (Well I actually read a bunch of const qualification code here and there and generally the latest version of that was quite easy to follow. I don't look at min_const_fn qualification though.^^)

I'm somewhat frustrated, TBH. I realize everyone is doing their best in the amount of time that they have. But also, given the amount of time I have, I just cannot read all that code -- if we cannot make this work based on design documents, I don't see how we can make it work at all. And I fear some of this will be really hard to sort out as it has been stable forever.

oli-obk:
>     * Inside a `static mut`, `&mut [...]` is also promoted. Why does that make any sense? Promoteds are consts I thought, they have read-only memory, don't they?

`static mut` allowed this from long before promotion, which is why we preserved it. We could start throwing future incompat warnings at ppl pushing them to use safe wrappers (see also https://github.com/rust-lang/rust/issues/53639)

>     * There's a special case [here](https://github.com/rust-lang/rust/blob/860d599d5b11db275a829a3f898560405fa84bbe/src/librustc_mir/transform/promote_consts.rs#L523) where I cannot figure out what it does.

this is a very recent change fixing a regression where a static wasn't able to promote its use of another static.



>     * I don't understand [this one](https://github.com/rust-lang/rust/blob/860d599d5b11db275a829a3f898560405fa84bbe/src/librustc_mir/transform/promote_consts.rs#L584) at all. Consts cannot refer to statics, sure, but promotion isn't the place to _enforce_ that property, or is it?

we carried this over from the original promotion impl which was also the const checker. We should look into just removing that check.



> Is there any chance we can ditch `const_kind` for promotion analysis, and make that all just depend on implicit vs explicit? That is at least a distinction that we have explored in some depth and we have some ideas for what the possible tradeoffs are.

Some of that mess is actually us trying to contain preexisting messes. We can defintely try to cut down on the variants of `ConstKind` by cratering things like the `static mut` stuff. I'm not sure how to remove the `*&STATIC` hack, since we try to keep mir building as simple as possible. But that's not actually using `const_kind`, so we can ignore it I guess?

RalfJung:
> static mut allowed this from long before promotion, which is why we preserved it. We could start throwing future incompat warnings at ppl pushing them to use safe wrappers (see also rust-lang/rust#53639)

That doesn't answer the question though -- if promoteds are evaluated like consts, won't this stuff be put into immutable memory? Or is there some other special handling elsewhere to make it mutable?

> this is a very recent change fixing a regression where a static wasn't able to promote its use of another static.

https://github.com/rust-lang/rust/pull/74945, I guess.
What about uses of statics elsewhere? Why would we only do this inside another static?
Looks like this is about e.g. `&(3, STATIC)` or so. Seems reasonable to promote everywhere?

> Some of that mess is actually us trying to contain preexisting messes.

Sure. But that doesn't explain why it is not described in documents like this one. The first step to cleaning up a mess is to properly understand the mess, and to me that means describing it in a higher level than code. The point of this document here was not to describe the ideal scenario, but to describe reality.

RalfJung:
I also just realized that our sanity check for explicit promotion, namely "if it fails to promote, it fails to compile", does *not* make sense for lifetime extension in `const`/`static` initializers. (I am so happy that we have this terminology now. :D ) So... the justification for using relaxed rules has to be a different one (and also the relaxed rules are different, but I am not sure if there is a good justification for that). This is in fact the same observation I already made in https://github.com/rust-lang/const-eval/issues/53 -- lifetime extension in `const`/`static` initializers is very special.

I see this as yet another argument that the old scheme of not doing lifetime extension via promotion, but simply by just dropping the local, actually made a lot of sense. The only thing that doesn't make sense is that this would be decided by a "promotion" analysis -- the concerns are very different, e.g. interior mutability would not be a problem I think.

oli-obk:
> Or is there some other special handling elsewhere to make it mutable?

I hope we're putting it in mutable memory in the interner, if not, we have an unsoundness and can fix it by just not promoting it anymore

> Seems reasonable to promote everywhere?

Yes, and we're doing that, we're just not allowing statics in constants, so we don't ever test this code path.

RalfJung:
> Yes, and we're doing that, we're just not allowing statics in constants, so we don't ever test this code path.

If I read the code correctly, we are not promoting `&(3, STATIC)` in a `fn`, as that would not enter the `if let Some(hir::ConstContext::Static(..)) = self.const_kind`.

RalfJung:
> I hope we're putting it in mutable memory in the interner, if not, we have an unsoundness and can fix it by just not promoting it anymore

The generated LLVM IR says it is indeed mutable. I just don't understand why, given [this line](https://github.com/rust-lang/rust/blob/8e5a27766ff14af36448a4b4c0a8bff84090168f/src/librustc_mir/interpret/intern.rs#L400).

oli-obk:
> I just don't understand why, given [this line](https://github.com/rust-lang/rust/blob/8e5a27766ff14af36448a4b4c0a8bff84090168f/src/librustc_mir/interpret/intern.rs#L400).

That surprised me, too, but I found the reason quickly: https://github.com/rust-lang/rust/blob/7189ca604ad5d9dfff7d0aeef6a42c89d73cbac7/src/librustc_mir/const_eval/eval_queries.rs#L61

RalfJung:
Not sure how that explains anything, isn't `static_mutability().is_some()` supposed to answer the question of "is this a static"?

oli-obk:
Yes, but for promoteds we just check their owner and not whether they are promoted unless it's not a static

RalfJung:
Could you point me to that code?

RalfJung:
> > Or is there some other special handling elsewhere to make it mutable?
> 
> I hope we're putting it in mutable memory in the interner, if not, we have an unsoundness and can fix it by just not promoting it anymore

Btw, not using promotion for lifetime extension of `&mut [...]` would also entirely bypass this problem in a backwards-compatible way. ;)

RalfJung:
Btw, after I raised my frustration earlier I should also say that the `promote_consts.rs` file actually is fairly pleasant to read, much better than its predecessor(s) -- I mean not in terms of all the mess that's encoded in there, but that's pre-existing, and there are some concrete checks I did not get, but the overall structure makes sense (after I realized how important `const_kind` is).

RalfJung:
> Yes, but for promoteds we just check their owner and not whether they are promoted unless it's not a static
> 
> So promoteds in statics are interned as statics

I only found [this code](https://github.com/rust-lang/rust/blob/7189ca604ad5d9dfff7d0aeef6a42c89d73cbac7/src/librustc_metadata/rmeta/decoder.rs#L1447) and it seems to require something with a `DefIndex` (is that the same as a `DefId`?), and not look into the parent of a promoted.

oli-obk:
The original thing I linked is what decides the interning mode: https://github.com/rust-lang/rust/blob/43bec40138bab10c08ac916bff2f2f81b2b70669/src/librustc_mir/const_eval/eval_queries.rs#L60-L70 So I'm not sure what else to link. This is the code that decides, and the match is only happening on the `DefId`, and the `DefId` of a promoted is the `DefId` of its owner.

RalfJung:
>  the DefId of a promoted is the DefId of its owner.

That is the part which I missed, thanks.

promotion.md
RalfJung: I have collected my current thoughts on this [in a hackmd](https://hackmd.io/7DekSpq3R-KgizegdAWO3Q). Please feel free to expand that with missing information!

RalfJung: I opened an issue at https://github.com/rust-lang/rust/issues/75586, so we have a dedicated place to track this problem. I will adjust the text here to reality.

RalfJung:
I updated the docs now to hopefully reflect reality. Could someone check that?

promotion.md
```diff
@@ -105,12 +95,14 @@ extension but actually does not:
...
 ```
 
-As we have seen above, `Vec::new()` does not get promoted. And yet this
+`Vec::new()` cannot get promoted because it needs dropping. And yet this
 compiles. Why that? The reason is that the reference obtains the lifetime of
 the "enclosing scope", similar to how `let x = &mut x;` creates a reference
 whose lifetime lasts for the enclosing scope. This is decided during MIR
 building already, and does not involve lifetime extension.
 
+This is also sometimes called "outermost scope" rule.
```
RalfJung, line 104: @eddyb calls it "outermost scope" rule. I am not sure why. I will try to find out.

RalfJung:
I updated these docs to reflect the current situation after the recent PRs that cleaned things up a bit.

PR #60, opened 2020-10-31
require CTFE to detect UB

RalfJung:
This is what @oli-obk and me agreed on [here](https://github.com/rust-lang/const-eval/pull/58#discussion_r508290645), except I went one step further and now require that for non-detected UB, CTFE continues in a reasonable way, instead of producing an arbitrary result. Once we keep the unoptimized MIR around, that is cheap to do. For example, this means that when the code violates alignment, CTFE must either raise an error or continue as if the pointer was actually well-aligned (and perform the appropriate load from the given region of memory). The previous wording would have allowed CTFE to return `Undef` when loading from an unaligned pointer -- I cannot see a good reason why we would permit CTFE to do that.

### DIFF
--- a/rfcs/0000-const-ub.md
+++ b/rfcs/0000-const-ub.md
@@ -45,7 +45,7 @@ Other kinds of UB might or might not be detected:
 * Dereferencing unaligned pointers.
 * Violating Rust's aliasing rules.
 * Producing an invalid value (but not using it in one of the ways defined above).
-* Anything [other UB][UB] not listed here.
+* Any [other UB][UB] not listed here.
 
 All of this UB has in common that there is an "obvious" way to continue evaluation even though the program has caused UB:
 we can just access the underlying memory despite alignment and/or aliasing rules being violated, and we can just ignore the existence of an invalid value as long as it is not used in some arithmetic, logical or control-flow operation.
--- a/rfcs/0000-const-ub.md
+++ b/rfcs/0000-const-ub.md
@@ -6,7 +6,8 @@
 # Summary
 [summary]: #summary
 
-Define UB during const evaluation to lead to an unspecified result for the affected CTFE query, but not otherwise infect the compilation process.
+Define how UB during const evaluation is treated:
+some kinds of UB must be detected, the rest leads to an unspecified result for the affected CTFE query (but does not otherwise "taint" the compilation process).
 
 # Motivation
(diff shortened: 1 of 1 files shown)


rfcs/0000-const-ub.md
```diff
@@ -21,43 +22,69 @@ There are some values that Rust needs to compute at compile-time.
...
-But other times, this might not be the case -- UB is not reliably detected during CTFE.
+The answer depends on the kind of UB: some kinds of UB are guaranteed to be detected,
+while other kinds of UB might either be detected, or else evaluation will continue as if the violated UB condition did not exist (i.e., as if this operation was actually defined).
 This can change from compiler version to compiler version: CTFE code that causes UB could build fine with one compiler and fail to build with another.
 (This is in accordance with the general policy that unsound code is not subject to strict stability guarantees.)
 
+[UB]: https://doc.rust-lang.org/reference/behavior-considered-undefined.html
+
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
```
RalfJung, line 35: As usual I have no idea how to split the RFC intention across "guide-level" and "reference-level" explanation.^^ That separation makes very little sense to me, for most of the RFCs that I end up writing...

rfcs/0000-const-ub.md
```diff
@@ -21,43 +22,69 @@ There are some values that Rust needs to compute at compile-time.
...
+* Dereferencing dangling pointers.
+* Using an invalid value in an arithmetic, logical or control-flow operation.
+
+These kinds of UB have in common that there is nothing sensible evaluation can do besides stopping with an error.
+
+Other kinds of UB might or might not be detected:
+* Dereferencing unaligned pointers.
+* Violating Rust's aliasing rules.
+* Producing an invalid value (but not using it in one of the ways defined above).
+* Anything [other UB][UB] not listed here.
```
oli-obk, line 48:
```suggestion
* Any [other UB][UB] not listed here.
```

oli-obk (approved):
> non-detected UB, CTFE continues in a reasonable way, instead of producing an arbitrary result. 

yes, we should not permit CTFE to explicitly create bogus values on UB when, in order to do that, it would need to detect the UB and choose to create a bogus value. If UB is detected, interpretation must be aborted. If UB is not detected, CTFE continues in a reasonable way.

PR #64, opened 2021-01-03
Update skill tree to current status

oli-obk:


src/skill_tree.md
```diff
@@ -72,22 +78,46 @@ href = "https://github.com/rust-lang/rust/issues/53020"
...
+label = "unconst operations"
 href = "https://github.com/rust-lang/const-eval/issues/14"
+items = []
+requires = ["unsafe_rules"]
+
+[[group]]
+name = "transmute"
+label = "feature:const_fn_transmute"
+items = []
+requires = ["unconst_rules"]
```
RalfJung, line 98: Why does "transmute" require "unconst" when the very similar union/raw-ptr require "unsafe"?

oli-obk: hmm overlooked that one

oli-obk: force pushed the last commit

RalfJung:
Thanks, looking good now. :)

PR #66, opened 2021-01-04
Update ptr::copy_nonoverlapping

usbalbin:
Update ptr::copy_nonoverlapping with feature gate and link to tracking issue.

### DIFF
--- a/src/skill_tree.md
+++ b/src/skill_tree.md
@@ -305,9 +305,10 @@ items = []
 requires = ["mut_ref", "trait_impl"]
 
 [[group]]
-label = "ptr::copy_nonoverlapping"
+label = "ptr::copy_nonoverlapping\nfeature:const_intrinsic_copy"
 name = "copy_nonoverlapping"
 items = []
+href = "https://github.com/rust-lang/rust/issues/80697"
 requires = ["raw_ptr_deref", "mut_ref"]
 
 [[group]]


usbalbin:
I have not quite figured out how to compile, I am currently getting
```
$ mdbook build
2021-01-04 19:57:22 [INFO] (mdbook::book): Book building has started
Warning: The mdbook-mermaid preprocessor was built against version 0.3.7 of mdbook, but we're being called from version 0.4.4
2021-01-04 19:57:22 [INFO] (mdbook::book): Running the html backend
2021-01-04 19:57:22 [ERROR] (mdbook::utils): Error: Rendering failed
2021-01-04 19:57:22 [ERROR] (mdbook::utils):    Caused By: Unable to copy across additional CSS and JS
2021-01-04 19:57:22 [ERROR] (mdbook::utils):    Caused By: Unable to copy /home/albin/rust/const-eval/skill-tree.css to /home/albin/rust/const-eval/book/skill-tree.css
2021-01-04 19:57:22 [ERROR] (mdbook::utils):    Caused By: No such file or directory (os error 2)
```

so my changes are not tested at all(!)

oli-obk:
Heh, yea, that failure is because the skill tree mdbook thing needs some initial setup. We haven't really documented anything, but in CI this is https://github.com/rust-lang/const-eval/blob/7f22c4ba54d2c37ca719297606c2dac0b1ec6c98/.github/workflows/deploy_skill_tree.yml#L32 (you need to install that command first)

oli-obk:
The change lgtm tho, so let's just make this CI's job so you don't have to install anything

usbalbin:
Ok, thanks :)

PR #67, opened 2021-05-09
update promotion docs

RalfJung:
"Explicit promotion" is no more :tada: , so update the docs accordingly.

Blocked on https://github.com/rust-lang/rust/pull/85110.

Fixes https://github.com/rust-lang/const-eval/issues/35
Fixes https://github.com/rust-lang/const-eval/issues/32

Fixes https://github.com/rust-lang/const-eval/issues/53

### DIFF
--- a/promotion.md
+++ b/promotion.md
@@ -2,80 +2,74 @@
 
 "Promotion" is the act of splicing a part of a MIR computation out into a
 separate self-contained MIR body which is evaluated at compile-time like a
-constant.
+constant. This mechanism has been introduced by [RFC 1414][promotion-rfc] with
+the goal of equipping some references-to-temporaries with a `'static` lifetime,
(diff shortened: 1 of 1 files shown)


RalfJung:
I marked this as also closing https://github.com/rust-lang/const-eval/issues/53. As far as I know, the only remaining "weirdness" of promotion is that `const fn` calls get promoted in `const`/`static` bodies, and this behavior is likely here to stay. It makes the entire system a bit less clean than I would like, but it also makes it more convenient for users, so maybe that's an okay trade-off. https://github.com/rust-lang/rust/pull/85112 ensures that rustc behaves correctly here.

RalfJung:
https://github.com/rust-lang/rust/pull/85110 landed. :)

PR #71, opened 2023-08-01
fix(promotion.md): remove broken link and related text

ismailarilik:
It seems that referred section doesn't exist.


promotion.md
```diff
@@ -71,9 +71,6 @@ are evaluated. We thus can be more relaxed about promotion; in practice, what
 this means is that we will promote calls to arbitrary `const fn`, not just those
 marked `#[rustc_promotable]`.
 
-[See below][static access] for another special case in promotion analysis:
```
oli-obk, line 74: It's called `access-static` now

ismailarilik: Done

PR #73, opened 2025-12-08
Fix minor typos in const-safety.md

jullanggit:
Considering this file hasn't been touched in seven years it might not be very relevant anymore, but in case it is, here are some minor typo fixes!


RalfJung:
Sure, why not -- thanks for the patch!

